PR #4, opened 2021-07-23
Add livenewsnow.com to the allowlist due to adblocker wall

jonathanKingston:
@jdorweiler FYI also.


SlayterDev:
This works but we should also add a `content-blocking-sites.json` for consistency.

SlayterDev (approved):

PR #6, opened 2021-07-26
Add content-blocking exceptions to legacy lists

SlayterDev:


jonathanKingston (approved):
Thanks for doing this!

PR #7, opened 2021-07-27
add tracker allowlist

jdorweiler:


exception-lists/tracker-allowlist.json
```diff
@@ -0,0 +1,4 @@
+{
```
SlayterDev, line 1: This file should follow the format of the other files in this directory. Namely the filename should be `tracker-allowlist-sites.json` and `allowlistedTrackers` should be an array.

SlayterDev (changes requested):

exception-lists/tracker-allowlist.json
jdorweiler: Why does it need to be an array?

SlayterDev: That's the current expected format. Changes will need to be made in `index.js` otherwise to support this and we should probably discuss it.

jdorweiler: Could it go under Settings instead?
```
trackerAllowlist: {
    settings: {
        allowlistedTrackers: {}
    }
}
```

SlayterDev (approved):
LGTM 👍

PR #9, opened 2021-07-29
Experiment with moving long term breakage to content blocking list

jonathanKingston:


jonathanKingston:
@sammacbeth this has only an impact to the extension currently so on par with your removals but perhaps less risk.

PR #16, opened 2021-08-10
Set version in build script as a timestamp

SlayterDev:


sammacbeth:
Are you sure that changing the type of this field from string to int won't cause issues in any of the implementations? iOS parsing seems to be particularly strict.

PR #28, opened 2021-09-06
Add pch.com to broken sites

jonathanKingston:
I'm adding this back due to the breakage numbers ongoing: https://app.asana.com/0/1200223097357040/1200913004760528/f

1. This is a contentBlocking change specifically for android which doesn't disable all other protections.
2. It merges all the platforms together for the various lists we consider to be part of the old global list.
3. I had to move the legacy list code to the bottom of the file, but is unchanged.
4. `www.sportskeeda.com` is also now in the legacy list  as it's in the `unprotectedTemporary` for iOS

### DIFF
--- a/generated/android-config.json
+++ b/generated/android-config.json
@@ -23,6 +23,10 @@
                 {
                     "domain": "id.me",
                     "reason": "Broken logins"
+                },
+                {
+                    "domain": "pch.com",
+                    "reason": "Broken site reports"
                 }
             ]
         },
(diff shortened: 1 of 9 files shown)


SlayterDev (approved):
LGTM 👍

SlayterDev:
There's some conflicts. If you resolve them go ahead and merge. @jonathanKingston

SlayterDev (approved):

PR #31, opened 2021-09-13
Move 53.com to content blocking list to improve privacy impact.

jonathanKingston:


exception-lists/content-blocking-sites.json
```diff
@@ -7,6 +7,10 @@
         }
     },
     "tempUnprotectedDomains": [
+        {
+            "domain": "53.com",
+            "reason": "site breakage"
```
kdzwinel, line 12: Can we be more explicit? Also, can we please link to the asana task from the description?

kdzwinel: I went ahead and added https://app.asana.com/0/1200223097357040/1200613477243552/f to the description. I also see that the breakage reason is unclear and untestable, but it seems to be linked to signing in? So maybe we can say "sign in issues".

jonathanKingston: Thanks! Rebased and updated.

kdzwinel (approved):
🧁

PR #32, opened 2021-09-13
Add costco.com to content blocking list.

jonathanKingston:


dharb (approved):
👍  thanks for this. I'll follow up on the fix to try to sort out what's actually causing the breakage.

ghostwords:
This one is strange. We get the 400 Bad Request page logging into Costco with Privacy Badger too. Here is what I got so far:

- Seems to be Chrome-specific. Doesn't seem to happen in Firefox, does happen in Edge (because Chromium).
- Seems to have started happening as of a recent Chrome update. I'm unable to reproduce in older Chrome versions. 
This suggests this could be a bug in Chromium.
- Bisecting Chromium builds produces [this changelog URL](https://chromium.googlesource.com/chromium/src/+log/c6e54aec9bf76c4e314e6100b7995975f9a23e99..8bb1537cc67b0fd50640aee16ce750b7412c9772)
- Privacy Badger sends both DNT and Sec-GPC headers by default. Sending DNT alone or Sec-GPC alone seems to work fine. Sending both consistently produces the error page. This explains why I couldn't reproduce with Brave (or Disconnect after enabling GPC). I'm not sure why DDG is affected though.

ghostwords:
Also, our user error reports for login problems on Costco only started in September for some reason.

ghostwords:
Perhaps it was the additional of yet another header in https://chromium.googlesource.com/chromium/src/+/ec5bce61388d15d34fa8d541e81d120ecaad9164. Costco's server has some kind of hardcoded header limit and our extensions happened to be one header away from it? Just thinking out loud.

PR #34, opened 2021-09-17
allow 3p cookies on bt.com

jdorweiler:
Cookie blocking breaks videos on bt.com


kdzwinel (approved):
LGTM although my understanding is that this won't work on apple platforms (we can't disable 3p cookies there).

PR #35, opened 2021-09-21
Linting and setup schema tests

SlayterDev:


.eslintrc.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
```
jonathanKingston, line 1: Can we make this use the .eslintrc file and structure to match the extension?

import.js
```diff
@@ -12,105 +12,105 @@ function getFile(path, host = 'duckduckgo.com') {
...
+    for (const legacyName in protectionsData) {
+        const legacyData = protectionsData[legacyName]
         // Known deprecated items we are removing
         if (legacyName === 'do-not-track') {
-            continue;
+            continue
         }
         if (!Object.values(legacyNaming).includes(legacyName)) {
-            console.debug("mapping name missing!", key);
+            console.debug('mapping name missing!', legacyName)
```
jonathanKingston, line 97: 👍🏻

tests/config-tests.js
```diff
@@ -0,0 +1,47 @@
+const expect = require('chai').expect
+const schema = require('./schema.js')
+const Ajv = require('ajv').default
+const ajv = new Ajv()
+const fs = require('fs')
+
+const platforms = [
```
jonathanKingston, line 7: Can this be made a shared import with index.js?

That way if we add a new platform we don't miss off running tests against it. Either that or make a helper that generates the array based on the files it finds.

tests/config-tests.js
```diff
@@ -0,0 +1,47 @@
+const expect = require('chai').expect
+const schema = require('./schema.js')
```
jonathanKingston, line 2: Could this be JSON instead of JS?

tests/schema.js
```diff
@@ -0,0 +1,30 @@
...
+        "properties": {
+            "readme": {"type": "string"},
+            "features": {"type": "object"},
+            "unprotectedTemporary": {
+                "type": "array",
+                "items": {"type": "object"}
+            }
+        }
+    },
+    "feature": {
```
jonathanKingston, line 13: Is there a way we can describe unique keys for each feature?

Perhaps in their own schema file perhaps? It doesn't need to be in this PR just interested on your thoughts where it would go.

tests/config-tests.js
```diff
@@ -0,0 +1,47 @@
...
+        for (const config of configs) {
+            for (const featureKey of Object.keys(config.features)) {
+                expect(ajv.validate(schema.feature, config.features[featureKey])).to.be.equal(true)
+            }
+        }
+    })
+
+    it('should have valid exception lists', () => {
+        for (const config of configs) {
+            for (const featureKey of Object.keys(config.features)) {
```
jonathanKingston, line 36:
```suggestion
            for (const featureKey in config.features) {
```

tests/config-tests.js
```diff
@@ -0,0 +1,47 @@
...
+describe('Data schema tests', () => {
+    it('should have a valid root schema', () => {
+        for (const config of configs) {
+            expect(ajv.validate(schema.root, config)).to.be.equal(true)
+        }
+    })
+
+    it('should have a vaild feature schema', () => {
+        for (const config of configs) {
+            for (const featureKey of Object.keys(config.features)) {
```
jonathanKingston, line 28:
```suggestion
            for (const featureKey in config.features) {
```

tests/schema.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
+    "root": {
```
jonathanKingston, line 2: Should these top level keys be unique files?

It's not clear looking at the file in isolation what is being checked.

Alternatively I think these can all be just joined together using referencing: https://ajv.js.org/guide/combining-schemas.html

jonathanKingston (commented):
Mostly just questions for you. Thanks for this :)

tests/schema.js
SlayterDev: Yeah we could define other schema files to match specific features.

tests/config-tests.js
SlayterDev: Node says `features` is not iterable.

tests/config-tests.js
SlayterDev: Same as above.

tests/config-tests.js
jonathanKingston: Are you using `in` instead of `of`?

jonathanKingston (approved):
It's shipping time

import.js
```diff
@@ -12,105 +12,105 @@ function getFile(path, host = 'duckduckgo.com') {
...
+        gpc: 'gpc',
+        autofill: 'autofill'
     }
-    function getNewName(legacyName) {
-        for (let newName in legacyNaming) {
-            if (legacyNaming[newName] == legacyName) {
-                return newName;
+    function getNewName (legacyName) {
+        for (const newName in legacyNaming) {
+            if (legacyNaming[newName] === legacyName) {
```
kdzwinel, line 75: oh hey, linting works!

index.js
```diff
@@ -139,19 +133,19 @@ const legacyTextDomains = [...unprotectedDomains].join('\n')
 fs.writeFileSync(`${GENERATED_DIR}/trackers-unprotected-temporary.txt`, legacyTextDomains)
 fs.writeFileSync(`${GENERATED_DIR}/trackers-whitelist-temporary.txt`, legacyTextDomains)
 const legacyNaming = {
```
kdzwinel, line 135: might be worth exporting as you did with `platforms.js` because it's both in this file and `import.js`

tests/schemas/root.json
```diff
@@ -0,0 +1,11 @@
+{
+    "type": "object",
+    "properties": {
+        "readme": {"type": "string"},
+        "features": {"type": "object"},
+        "unprotectedTemporary": {
+            "type": "array",
+            "items": {"type": "object"}
+        }
```
kdzwinel, line 9: what about "version"? BTW can we make validation warn us about unexpected fields?

tests/schemas/feature.json
```diff
@@ -0,0 +1,10 @@
+{
+    "type": "object",
+    "properties": {
+        "state": {"type": "string"},
```
kdzwinel, line 4: there are only few valid states, can we have an enum here instead of a string?

kdzwinel (commented):
I'm late to the party, but I left some ideas for improvement!

tests/schemas/feature.json
kdzwinel: Looks like it's doable:

> "default": {"type": "string", "enum": ["block", "ignore"]},

(looking at Jason's work here: https://dub.duckduckgo.com/duckduckgo/tracker-radar-internal/blob/master/tests/schema.js )

PR #37, opened 2021-09-23
Setup for automatic building

SlayterDev:
This PR sets us up to build the configs on merge. This is done using the following process:

When a PR is merged to main:
- Checkout the repo
- Run `index.js`
- Publish the built files to S3
- Commit the newly built files to main


index.js
```diff
@@ -5,7 +5,10 @@ const GENERATED_DIR = 'generated'
 const LISTS_DIR = 'exception-lists'
 
 const defaultConfig = JSON.parse(fs.readFileSync('default-config.json'))
-defaultConfig.version = Date.now()
+
+if (process.env.USING_CI) {
```
jonathanKingston, line 9: Is this needed?

1. Are we envisioning this to still be run by config authors?
2. Should we remove the `generated/` directory from the repo and also add it to the `.gitignore`?

.github/workflows/build-and-publish.yml
```diff
@@ -19,3 +23,9 @@ jobs:
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         SOURCE_DIR: 'generated' 
         DEST_DIR: 'trackerblocking/config/v1'
+    - uses: EndBug/add-and-commit@v7
```
jonathanKingston, line 26: I'm a bit of a beginner at reading these files (we should probably have an additional reviewer that has done more of it like @GioSensation), are we actually calling `node index.js` or `npm start` or similar here? Also if the unit tests just added fail, would this break before uploading to s3?

jonathanKingston (commented):
Thanks!
Some questions and the suggestion to get another reviewer IE @GioSensation or similar :).

.github/workflows/build-and-publish.yml
SlayterDev: This action only runs _after_ merge so tests should be passing at this stage. Also `line 16` is what runs `index.js` via [this file](https://github.com/duckduckgo/privacy-configuration/blob/main/action.yml)

index.js
SlayterDev: This is up for debate. When we previously ran the build script in the action we had the issue that a maintainer would:
- generate locally creating one timestamp
- Merge their PR which builds and generates another timestamp

This would cause confusion because the timestamp in the repo was different than the published version.

I'm personally in favor of keeping the generated files in the repo as a quick reference of what's in prod however @kdzwinel also voiced that we could ignore these files in the repo and only build in the actions and just use the online versions for reference (possibly allowing s3 to list the directory their in).

jonathanKingston: I think if we want full visibility of published code we could have the CI create a PR to a different repo.

The rationale for removing the generated changes is that it'll reduce the amount of merge conflicts and noise for a reviewer.

Perhaps we can mark the generated files as binaries? Would this reduce the noise in github / git?

.github/workflows/build-and-publish.yml
jonathanKingston: I think ideally it should also run tests just in case a broken merge gets put into main?

index.js
kdzwinel: Uuu… it'd be fancy to release generated files to https://github.com/duckduckgo/privacy-configuration/releases . This action seems to do that: https://github.com/marvinpinto/action-automatic-releases

SlayterDev: Wish granted 🪄

.github/workflows/build-and-publish.yml
```diff
@@ -19,3 +24,12 @@ jobs:
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         SOURCE_DIR: 'generated' 
         DEST_DIR: 'trackerblocking/config/v1'
+    - name: Get current date
+      id: date
+      run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
```
jonathanKingston, line 29: Does this get cached? Should we have hour here?

Alternatively we could take it from the generated version in the config?

jonathanKingston:
LGTM
@giosensation or @kdzwinel are you able to double check this as I believe you have some experience with workflows.

.github/workflows/build-and-publish.yml
```diff
@@ -10,6 +10,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@main
+    - name: Install dependencies
+      run: npm install
+    - uses: ./
+    - name: Run tests
+      run: npm run test
```
kdzwinel, line 17: will whole action fail if tests fail?

SlayterDev: This is my understanding

kdzwinel: we should test that once we merge this, or on some test repo if we want to test it before

SlayterDev: I can't find anything that definitively states this but I think the default behavior of actions is to exit on failure

GioSensation: Yes, actions fail if a script returns something other than 0 or throws.

GioSensation (approved):
LGTM.

jonathanKingston (approved):

PR #47, opened 2021-10-16
Play around with moving the file structure to simplify config mainten…

jonathanKingston:
…ance

This is just a draft idea, I think everything is working as expected but made it to explain the task alternative in: https://app.asana.com/0/1200890834746050/1201208544801855/f

To verify the output here:

- git checkout main
- node index.js
- cp generated generated.bk
- jq --sort-keys . generated.bk/extension-config.json > generated.bk/extension-config.json.sorted
- git checkout jkt/list-change
- node index.js
- jq --sort-keys . generated/extension-config.json > generated/extension-config.json.sorted
- diff generated.bk/extension-config.json.sorted generated/extension-config.json.sorted

Expected outcomes:

- Modified items should only be: version
- Removed features should only be where "state" is not "enabled"


index.js
```diff
@@ -95,6 +84,10 @@ for (const platform of platforms) {
                 platformConfig.features[key].exceptions = platformConfig.features[key].exceptions.concat(platformOverride.features[key].exceptions)
             }
         }
+        if (platformConfig.features[key].state !== 'enabled') {
```
SlayterDev, line 87: This doesn't seem future proof. Especially if we add features in beta/experimental state. I would think it would be better to check specifically for `disabled` if we're set on removing them from the config.

SlayterDev (changes requested):

index.js
jonathanKingston: None of the code caters for other states either, there's a much bigger project there.

I can make the change for id there's no state or "disabled" as mentioned though.

SlayterDev:
@jonathanKingston can you handle the conflicts here and I think we can get this in.

jonathanKingston:
> @jonathanKingston can you handle the conflicts here and I think we can get this in.

Certainly, I'll do one last diff to ensure the output is identical too :)

jonathanKingston:
@SlayterDev only the expected changed were visible in the output like disabled features not being present and versioning. Everything is rebased and working now :)

SlayterDev (approved):

PR #66, opened 2021-11-03
Add tracking links feature to iOS

SlayterDev:
**Task:** https://app.asana.com/0/72649045549333/1201286795398788/f

This is a draft for now but I'll use this feature to serve two types of AMP formats and a list of url tracking parameters:
- AMP links we can extract the real URL from
- AMP links we have to visit first
- FB/Google url parameters used for tracking.


jonathanKingston (approved):
LGTM thanks for looking into the regex thing :)

PR #70, opened 2021-11-05
Remove Costco.com

jonathanKingston:
Based on feedback here: https://github.com/duckduckgo/privacy-configuration/pull/32 and continued breakage reports despite this


GuiltyDolphin:
We should remove this from the generated files too?

SlayterDev (approved):

jonathanKingston:
> We should remove this from the generated files too?

It's dead jim :D

We removed this as it's no longer needed again.

PR #74, opened 2021-11-09
Remove disabled pruning

jonathanKingston:
For https://github.com/duckduckgo/duckduckgo-privacy-extension/issues/929


kdzwinel (approved):
Let's clean up those if's as a follow up

PR #75, opened 2021-11-09
Clean up removed code

jonathanKingston:


index.js
```diff
@@ -135,7 +127,7 @@ const legacyNaming = {
 const protections = {}
 for (const key in legacyNaming) {
     let newConfig
-    if (!defaultConfig.features[key] || defaultConfig.features[key].state !== 'enabled') {
```
kdzwinel, line 138: so previously for things like 'beta' (that we don't use just yet AFAIK) this would be true and now it's false, is that expected?

index.js
```diff
@@ -88,10 +88,6 @@ for (const platform of platforms) {
                 platformConfig.features[key].exceptions = platformConfig.features[key].exceptions.concat(platformOverride.features[key].exceptions)
             }
         }
-        if (isFeatureDisabled(platformConfig.features[key])) {
```
kdzwinel, line 91: hm, I thought we commented out this code out because missing feature keys were causing issues?

kdzwinel (commented):
sorry for late review, I'm probably not the best person to review anyway because I don't know this code. I left some comments, but not sure if those are helpful

index.js
jonathanKingston: Just needed a rebase, was done at the same time github wasn't helping as it was a clean rebase.

index.js
jonathanKingston: I don't think we should be coding for this beta stuff as there's no supporting code. The format if forwards compatible and that's all we needed.


I've removed all this code as now we've simplified things we can just simplify the transform here.

index.js
```diff
@@ -88,9 +90,7 @@ for (const platform of platforms) {
...
-            // If feature isn't enabled for platform remove.
-            // delete platformConfig.features[key]
+        if (isFeatureMissingState(platformConfig.features[key])) {
```
kdzwinel, line 93: Sorry, I'm still not the best person to review, but I was wondering - why would feature miss a state in the first place? Why don't we enforce state to always be there via JSON schema? This way we will force folks to directly declare it and don't make any assumptions.

index.js
kdzwinel: Yup, fair point about the beta 👍

index.js
jonathanKingston: A feature might only be enabled on one platform.
The previous approach was to delete these keys but there's at least an issue in the extension currently doing this. In the new year we should be good to delete again.

PR #79, opened 2021-11-11
Match feature state to feature table

SlayterDev:
https://app.asana.com/0/1200890834746050/1201349025761436/f

I used https://app.asana.com/0/1200890834746050/1201287926785314/f as a guide to match which features are only on certain platforms. The generated configs should properly reflect the supported feature set. Some of our newer features (trackingLinks, navigatorCredentials) were moved into the `features` dir.

Also I think doing this broke something slightly with generating legacy configs. Namely the `protections.json` file. Please check it works as expected.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -3,5 +3,5 @@ description: 'Build generated privacy configs'
 author: DuckDuckGo
 
 runs:
-  using: 'node12'
+  using: 'node16'
   main: 'index.js'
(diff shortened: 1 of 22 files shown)


jonathanKingston:
Yeah looks like the merging of legacy configs isn't working correctly here. We should generate this based on the combined output of the extension config, which likely simplifies this code.

- I don't think we should land this without fixing it. Specifically "canvas" and "hardware" should contain the exception lists.

After running jq on the generated outputs everything else appears to be in order.

jonathanKingston (approved):

jonathanKingston:
LGTM thanks for fixing the legacy part!

PR #80, opened 2021-11-11
Update schema tests

SlayterDev:
https://app.asana.com/0/1200890834746050/1201359375306447/f


tests/schemas/feature.json
```diff
@@ -1,10 +1,15 @@
 {
     "type": "object",
     "properties": {
-        "state": {"type": "string"},
+        "state": {"enum": ["enabled", "disabled", "beta"]},
         "exceptions": {
             "type": "array",
             "items": {"type": "object"}
+        },
+        "settings": {
```
kdzwinel, line 9: if we specify `settings` as optional field then we can say `additionalProperties: false`

tests/schemas/root.json
```diff
@@ -2,10 +2,19 @@
     "type": "object",
     "properties": {
         "readme": {"type": "string"},
-        "features": {"type": "object"},
+        "features": {
+            "type": "object",
+            "patternProperties": {
+                "^[a-zA-Z0-9]+$": {"type": "object"}
```
kdzwinel, line 8: this makes sure that feature name is using only [a-zA-Z0-9] in its name. We could paste whole `feature.json` in here then, but I'm OK with keeping it separate

tests/config-tests.js
```diff
@@ -4,39 +4,52 @@ const ajv = new Ajv()
 const fs = require('fs')
 const platforms = require('./../platforms')
 
+function formatErrors (errors) {
+    if (!Array.isArray(errors)) {
+        return ''
+    }
+
+    return errors.map(item => `${item.instancePath}: ${item.message}`).join(', ')
```
kdzwinel, line 12: `ajv` returns some error messages that make it easier to track down bugs
more-info

tests/config-tests.js
```diff
@@ -4,39 +4,52 @@ const ajv = new Ajv()
...
-                for (const exception of config.features[featureKey].exceptions) {
-                    expect(ajv.validate(schema, exception)).to.be.equal(true)
+    const rootSchema = JSON.parse(fs.readFileSync('./tests/schemas/root.json'))
+    const validateRoot = ajv.compile(rootSchema)
+    const featureSchema = JSON.parse(fs.readFileSync('./tests/schemas/feature.json'))
+    const validateFeature = ajv.compile(featureSchema)
+    const exceptionSchema = JSON.parse(fs.readFileSync('./tests/schemas/exception.json'))
+    const validateException = ajv.compile(exceptionSchema)
+
+    for (const config of configs) {
```
kdzwinel, line 30: this way, we make it easier to find where the bug is and iterate over list of configs only once. The output looks like this: 

success

kdzwinel (commented):
Thanks Brad, this is great! I improved it a bit more by making it easier to find errors and making feature checking even more strict.

kdzwinel (approved):

PR #87, opened 2021-11-19
Split tracking link features

SlayterDev:
Split the tracking links feature into two features.

Asana context: https://app.asana.com/0/0/1201299313936485/1201380456152563/f

### DIFF
--- a/overrides/ios-override.json
+++ b/overrides/ios-override.json
@@ -12,11 +12,11 @@
             "state": "enabled",
             "exceptions": [],
             "settings": {
-                "ampLinkFormats": [
+                "linkFormats": [
                     "https?:\\/\\/(?:w{3}\\.)?google\\.\\w{2,}\\/amp\\/s\\/(\\S+)",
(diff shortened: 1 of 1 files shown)


overrides/ios-override.json
```diff
@@ -30,7 +30,13 @@
...
                     "?amp"
-                ],
+                ]
+            }
+        },
+        "trackingParameters": {
+            "state": "enabled",
+            "exceptions": [],
+            "settings": {
                 "trackingParameters": [
```
jonathanKingston, line 40: Can we just call this parameters given the feature name?

overrides/ios-override.json
```diff
@@ -8,7 +8,7 @@
                 }
             ]
         },
-        "trackingLinks": {
+        "ampLinks": {
             "state": "enabled",
             "exceptions": [],
             "settings": {
```
jonathanKingston, line 14: Given the feature name can we remove the 'amp' prefix on the 'settings'?

jonathanKingston (approved):

PR #102, opened 2022-01-03
Exception for samuelschmitt.com

SlayterDev:
Breakage discovered here: https://app.asana.com/0/1142021229838617/1201477006467551/f

### DIFF
--- a/features/amp-links.json
+++ b/features/amp-links.json
@@ -6,7 +6,12 @@
             "reason": "site breakage"
         }
     },
-    "exceptions": [],
+    "exceptions": [
+        {
+            "domain": "samuelschmitt.com",
+            "reason": "Links rewrite to invailid pages"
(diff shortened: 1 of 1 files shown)


jonathanKingston:
This is position 222489 on the tranco list, so we wouldn't classify this as critical breakage.

Is there an inherent problem here that could be on other websites?

Maybe tracking this in Asana as it's own issue to discuss is worthwhile.

SlayterDev:
Honestly I'm fine with not putting in the exception given its so small was just following up with a discovered issue.

As far as tracking the root cause in Asana we are discussing it. However due to the nature of AMP links there may not be a fix for this one rather specific issue.

GuiltyDolphin:
Closing this because it's not critical breakgage (not in our tranco or standard lists).

PR #115, opened 2022-01-27
Enable browsingTopics on extension config

SlayterDev:
https://app.asana.com/0/1163321984198618/1201719300245228/f

`browsingTopics` is effectively replacing `floc`. This PR enables the browsingTopics protection for the extension.

**Note:** I've kept floc enabled in the config for previous extension versions.

### DIFF
--- a/index.js
+++ b/index.js
@@ -45,6 +45,12 @@ function generateV1Config (platformConfig) {
             delete v1Config.features[feature].minSupportedVersion
             v1Config.features[feature].state = 'disabled'
         }
+
+        if (v1Config.features[feature].eol === 'v1') {
+            // This feature's support ends in v1 so remove it from platformConfig
+            delete v1Config.features[feature].eol
+            delete platformConfig.features[feature]
+        }
     }
 
     return v1Config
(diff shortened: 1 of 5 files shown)


overrides/extension-override.json
```diff
@@ -33,6 +33,9 @@
         "fingerprintingBattery": {
             "state": "enabled"
         },
+        "browsingTopics": {
+            "state": "enabled"
+        },
         "floc": {
             "state": "enabled"
```
jonathanKingston, line 40: Could we ditch floc in v1?

index.js
```diff
@@ -45,6 +45,12 @@ function generateV1Config (platformConfig) {
             delete v1Config.features[feature].minSupportedVersion
             v1Config.features[feature].state = 'disabled'
         }
+
+        if (v1Config.features[feature].eol === 'v1') {
```
jonathanKingston, line 49: Seems fine but we should genericise this in the improvements patch.

jonathanKingston (approved):
SGTM update and rebase etc. Happy to have this merged before the final cut of the extension/content-scope change.

PR #122, opened 2022-02-02
Push branch to staging

SlayterDev:
https://app.asana.com/0/1200890834746050/1201724156469596/f

### DIFF
--- a/.github/workflows/push-to-stage.yml
+++ b/.github/workflows/push-to-stage.yml
@@ -1,6 +1,10 @@
 name: Push to Staging
 
-on: [push, pull_request]
+on: 
+  push:
+  pull_request:
+    types:
+      - opened
 
 jobs:
   publish:
@@ -25,3 +29,9 @@ jobs:
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         SOURCE_DIR: 'generated' 
         DEST_DIR: 'trackerblocking/config/staging/${{ steps.extract_branch.outputs.branch }}'
+    - uses: github-actions-up-and-running/pr-comment@v1.0.0
+      if: github.event.action == 'opened'
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        message: "Your PR is being temporarily hosted at https://staticcdn.duckduckgo.com/trackerblocking/config/staging/${{ steps.extract_branch.outputs.branch }} until this branch is deleted."
+
--- a/.github/workflows/push-to-stage.yml
+++ b/.github/workflows/push-to-stage.yml
@@ -18,7 +18,7 @@ jobs:
     - name: Extract branch name (PR Only)
       shell: bash
       run: echo "##[set-output name=branch;]$(echo ${GITHUB_HEAD_REF#refs/heads/})"
-      id: extract_branch
+      id: extract_branch_pr
     - uses: actions/checkout@v2
     - name: Install dependencies
       run: npm install
@@ -39,5 +39,5 @@ jobs:
       if: github.event.action == 'opened'
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        message: 'Your PR is being temporarily hosted at https://staticcdn.duckduckgo.com/trackerblocking/config/staging/${{ steps.extract_branch.outputs.branch }} until this branch is deleted.'
+        message: 'Your PR is being temporarily hosted at https://staticcdn.duckduckgo.com/trackerblocking/config/staging/${{ steps.extract_branch_pr.outputs.branch }} until this branch is deleted.'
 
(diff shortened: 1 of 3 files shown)


.github/workflows/push-to-stage.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Extract branch name
+      shell: bash
+      run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+      id: extract_branch
+    # Github uses different branch var names for push vs PR event types
+    - name: Extract branch name (PR Only)
+      if: github.event.action == 'opened'
+      shell: bash
+      run: echo "##[set-output name=branch;]$(echo ${GITHUB_HEAD_REF#refs/heads/})"
+      id: extract_branch_pr
```
sammacbeth, line 22: Will this work for cross-fork PRs? What if the same branch name exists on multiple forks with open PRs? Could we not use something guaranteed to be mutually exclusive, like PR id, instead?

.github/workflows/push-to-stage.yml
```diff
@@ -0,0 +1,44 @@
+name: Push to Staging
+
+on: 
+  push:
+  pull_request:
+    types:
+      - opened
```
sammacbeth, line 7: Will this rerun if I push new commits to my fork branch which already has an open PR?

.github/workflows/push-to-stage.yml
```diff
@@ -0,0 +1,44 @@
...
+    - uses: jakejarvis/s3-sync-action@master
+      if: github.event_name == 'push'
+      with:
+        args: --acl public-read --follow-symlinks
+      env:
+        AWS_S3_BUCKET: ${{ secrets.AWS_S3_BUCKET }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        SOURCE_DIR: 'generated' 
+        DEST_DIR: 'trackerblocking/config/staging/${{ steps.extract_branch.outputs.branch }}'
```
sammacbeth, line 38: Is the output of the `extract_branch_pr` step ever used?

sammacbeth (commented):
I added some comments.

I also wonder if we need some protection for who can trigger this workflow. As this is a public repo anyone could open a PR and publish arbitrary files to our CDN. That doesn't sound like a good idea.

.github/workflows/push-to-stage.yml
SlayterDev: The pull request event will only fire when the PR is opened. In this workflow we also fire on push to update the config files. The pr opened event is mainly listened for to add the comment with the config url.

.github/workflows/push-to-stage.yml
SlayterDev: Yes. There's a step to add a comment to the PR with the hosted config location. That's where that output is used. It's done on the PR opened event so it only happens once.

.github/workflows/push-to-stage.yml
SlayterDev: As discussed outside contributors won't be able to run these workflows so I believe this is a non issue re: branch name conflicts.

.github/workflows/remove-from-stage.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  remove:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Extract branch name
+      shell: bash
+      run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+      id: extract_branch
+    - name: Remove from s3
+      uses: vitorsgomes/s3-rm-action@master
```
sammacbeth, line 18: Please pin 3rd party actions to a commit hash. Otherwise a compromise of this repo could gain control of our AWS access keys.

.github/workflows/push-to-stage.yml
```diff
@@ -0,0 +1,44 @@
...
+      shell: bash
+      run: echo "##[set-output name=branch;]$(echo ${GITHUB_HEAD_REF#refs/heads/})"
+      id: extract_branch_pr
+    - uses: actions/checkout@v2
+    - name: Install dependencies
+      run: npm install
+    - uses: ./
+    - name: Run tests
+      run: npm run test
+    - uses: jakejarvis/s3-sync-action@master
```
sammacbeth, line 29: Please pin this action to a commit.

.github/workflows/remove-from-stage.yml
```diff
@@ -0,0 +1,25 @@
...
+      with:
+        args: --recursive
+      env:
+        AWS_S3_BUCKET: ${{ secrets.AWS_S3_BUCKET }}
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        PATH_TO_DELETE: 'trackerblocking/config/staging/${{ steps.extract_branch.outputs.branch }}'
```
sammacbeth, line 25: Is this a prefix match or path? I wonder if there exists a branch `sam` and `samv2` and I delete `sam` if this action might delete the files in `samv2` too. I think a trailing slash would resolve it. You don't have to delete the directory on s3 as it simply ceases to exist once all the files in it are deleted.

SlayterDev: I believe its a path but I'll add the trailing slash to eliminate ambiguity.

sammacbeth (approved):

PR #137, opened 2022-02-14
Disable AMP Link processing on Android

joshliebe:
We're seeing an increase in `WebView` crashes, so we're going to temporarily disable this feature to see if it is the cause.

Task: https://app.asana.com/0/414730916066338/1201814461914604/f


marcosholgado (approved):
LGTM

kdzwinel (approved):
Lgtm

PR #140, opened 2022-02-21
Enable AMP Link processing on Android

joshliebe:
We are confident that this was not the cause of the `WebView` crashes in the previous Android release, so we are re-enabling the feature.
I will continue to monitor once the feature is enabled.

Task: https://app.asana.com/0/414730916066338/1201850932614952/f

### DIFF
--- a/overrides/android-override.json
+++ b/overrides/android-override.json
@@ -26,7 +26,7 @@
             ]
         },
         "ampLinks": {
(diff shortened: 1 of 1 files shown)


SlayterDev (commented):
LGTM 👍

SlayterDev (approved):

PR #162, opened 2022-03-23
Disable tracking parameters feature

joshliebe:
There is a crash when opening a parameter link from a mail in gmail.com (At least that's the only way I could reproduce it), so I'm disabling this feature until the issue is fixed.

Task: https://app.asana.com/0/414730916066338/1202011815553692/f


kdzwinel:
LGTM

PR #176, opened 2022-03-31
Enable v2 endpoint and handle minSupportedVersion

SlayterDev:
https://app.asana.com/0/1200890834746050/1201683113648489/f

This PR pushes config to v1 and v2 directories. If a feature contains the `minSupportedVersion` key we'll delete this key from v1 and disable the feature.

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,7 +24,7 @@ function writeConfigToDisk (platform, config, v1Config) {
 }
 
 /**
- * Create the specified directory if it doesn't exist 
+ * Create the specified directory if it doesn't exist
  *
  * @param {string} dir - directory path to create
  */
--- a/index.js
+++ b/index.js
@@ -18,8 +18,25 @@ const platforms = require('./platforms')
  * @param {string} platform - platform to write
  * @param {object} config - the object to write
  */
-function writeConfigToDisk (platform, config) {
-    fs.writeFileSync(`${GENERATED_DIR}/${platform}-config.json`, JSON.stringify(config, null, 4))
+function writeConfigToDisk (platform, config, v1Config) {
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -52,6 +69,10 @@ addExceptionsToUnprotected(defaultConfig.features.contentBlocking.exceptions)
 
 if (!fs.existsSync(GENERATED_DIR)) {
     fs.mkdirSync(GENERATED_DIR)
+
+    // Create version directories
+    fs.mkdirSync(`${GENERATED_DIR}/v1`)
+    fs.mkdirSync(`${GENERATED_DIR}/v2`)
```
jonathanKingston, line 75: Can we make the `existsSync` and `mkdirSync` a function and call it for all three directories? Locally this broke for me as I already had a generated/ but not the other dirs. Ideally users of the config shouldn't need to understand about the generation process.


Also worth noting this path is mentioned in the extensions CONTRIBUTING.md for local development.  (happy to add this to the task of integrating the update to v2)
Perhaps other projects are depending on the generated directory being the route?

tests/schemas/feature.json
```diff
@@ -8,7 +8,8 @@
         },
         "settings": {
             "type": "object"
-        }
+        },
+        "minSupportedVersion": {"type": "string"}
```
jonathanKingston, line 12: 👍🏻

index.js
SlayterDev: I'll get those into a function.

Regarding the change of local path yeah that's something we should probably delegate to v2 integration projects.

jonathanKingston (approved):
LGTM if you're merging it ensure the output of the generated build matches before and after merge please :)

PR #193, opened 2022-04-07
Update ampproject regex

SlayterDev:
AMP url not previously matched by this regex: https://www-usatoday-com.cdn.ampproject.org/c/s/www.usatoday.com/amp-stories/beyonce-rule-the-world/


jonathanKingston (approved):
Seems good to me.

I think we should continue the discussion of these formats in: https://app.asana.com/0/1200890834746050/1202119310370827/f as I think we're at risk of introducing bugs here.

PR #194, opened 2022-04-07
Zip generated files before creating release

SlayterDev:
This will place all the files in the generated dir into a zip archive which will be used for releases instead of the individual json files.


jonathanKingston (approved):
SGTM rebase and squash merge :)

PR #195, opened 2022-04-07
Initial project restructure

SlayterDev:
Took a stab at restructuring the generation scripts a bit. Now that we're getting more complex it will be helpful to modularize some areas.


index.js
```diff
@@ -1,102 +1,53 @@
 const fs = require('fs')
-
-const OVERRIDE_DIR = 'overrides'
-const GENERATED_DIR = 'generated'
-const LISTS_DIR = 'features'
+const utils = require('./lib/utils')
```
jonathanKingston, line 2: This brings in a whole heap of require/export can we move to ESM like content-scope-scripts does? It should be just a case of adding _"type": "module"_ to the package.json as we're just a node script.

lib/constants.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+    OVERRIDE_DIR: 'overrides',
```
jonathanKingston, line 2: Moving to a constants file makes sense when they're used in multiple places, I think they're only in index.js right?

index.js
```diff
@@ -1,102 +1,53 @@
...
 
 // Handle platform specific overrides and write configs to disk
 for (const platform of platforms) {
     const platformConfig = JSON.parse(JSON.stringify(defaultConfig))
-    const overridePath = `${OVERRIDE_DIR}/${platform}-override.json`
+    const overridePath = `${constants.OVERRIDE_DIR}/${platform}-override.json`
 
     if (!fs.existsSync(overridePath)) {
-        writeConfigToDisk(platform, platformConfig)
+        utils.writeConfigToDisk(platform, platformConfig)
```
jonathanKingston, line 50: This seems like a *existing* bug, we're not passing the second argument and we're creating the generated config files if the override doesn't exist?!
Probably can be deleted.

lib/utils.js
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Write a config file to disk
+ *
+ * @param {string} platform - platform to write
+ * @param {object} config - the object to write
+ */
+function writeConfigToDisk (platform, config, v1Config) {
+    fs.writeFileSync(`${constants.GENERATED_DIR}/v2/${platform}-config.json`, JSON.stringify(config, null, 4))
+    fs.writeFileSync(`${constants.GENERATED_DIR}/v1/${platform}-config.json`, JSON.stringify(v1Config, null, 4))
```
jonathanKingston, line 12: This v1 saving would be better as part of the legacy generation, in hindsight the `v1Config` argument here seems like a mistake...

lib/legacy-gen/v1generator.js
```diff
@@ -0,0 +1,19 @@
+function generateV1Config (platformConfig) {
```
jonathanKingston, line 1: Both this and legacy could be implemented as transitions on the current format such that the code would just recursively call it and then there'd be less versioned code over time. v0 could special case the old files perhaps too.

```js
export const legacyFormats = [
   {
        configVersion: 'v1',
        transition(dataIn) {
            const dataOut = dataIn;
            ...
            return dataOut
        },
    }
    {
        configVersion: 'v0',
        ....
    }
]
```

```js
import { legacyFormats } from './legacyFormats.js'

function writeToDisk(currentFormat) {
    writeConfigToDisk(currentConfigVersion, platform, dataOut)
    for (let format of legacyFormats) {
        const dataOut = format.transition(currentFormat)
        if (format.configVersion === 'v0') {
            writeLegacyFilesToDisk(dataOut)
        } else {
            writeConfigToDisk(format.configVersion, platform, dataOut)
        }
        currentFormat = dataOut
    }
}
```

jonathanKingston (changes requested):
I think there are bigger changes we can bring within this that are worth discussing... added my comments.

lib/constants.js
SlayterDev: If we move to saving in the legacy generation they'll be used elsewhere. I think at least GENERATED is already used in a couple files.

lib/legacy-gen/v1generator.js
SlayterDev: I like the idea of the legacy transition structure but I think the v0 config files are just special enough, especially with the need for the unprotectedTemp object, that they should be kept as a separate case done at then end of generation. The others make sense to be done per platform file on write but the v0 files only need to be done once.

jonathanKingston:
Let's add the genericising of the eol feature here too. Otherwise this looks good; I'll wait for the fledge change to drop before giving the final r+ here.

SlayterDev:
Closing for now as this will need some refactoring after recent updates.

PR #215, opened 2022-04-14
Add containerstore to autofill exception on macos

GioSensation:
This site has multiple autofill issues. See [Asana](https://app.asana.com/0/1200930669568058/1202105131841960/f) for context.

### DIFF
--- a/overrides/macos-override.json
+++ b/overrides/macos-override.json
@@ -35,7 +35,7 @@
             "exceptions": [
                 {
                     "domain": "containerstore.com",
-                    "reason": "Poor UX: https://app.asana.com/0/1200930669568058/1202105131841960/f"
+                    "reason": "Poor UX"
                 }
             ]
         },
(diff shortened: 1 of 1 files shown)


GioSensation:
@jonathanKingston I've added this exception for autofill on macOS. Is this the right way to do it?

overrides/macos-override.json
```diff
@@ -31,7 +31,13 @@
             "state": "enabled"
         },
         "autofill": {
-            "state": "enabled"
+            "state": "enabled",
+            "exceptions": [
+                {
+                    "domain": "containerstore.com",
+                    "reason": "Poor UX: https://app.asana.com/0/1200930669568058/1202105131841960/f"
```
SlayterDev, line 38: Let's keep Asana links out of the config. There is potential one day in the future these reasons could be user facing.

SlayterDev (changes requested):

overrides/macos-override.json
GioSensation: Done!

SlayterDev (approved):
LGTM 👍

PR #216, opened 2022-04-14
Re-enable tracking parameters protection on mac

SlayterDev:
https://app.asana.com/0/414235014887631/1202080405650495/f has shipped so we can re-enable.

### DIFF
--- a/overrides/macos-override.json
+++ b/overrides/macos-override.json
@@ -43,7 +43,7 @@
             "state": "enabled"
         },
         "trackingParameters": {
-            "state": "disabled"
+            "state": "enabled"
         }
     },
     "unprotectedTemporary": [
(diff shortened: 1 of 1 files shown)


jonathanKingston:
Can we make this v2 only to prevent breakage?

overrides/macos-override.json
```diff
@@ -43,7 +43,8 @@
             "state": "enabled"
         },
         "trackingParameters": {
-            "state": "disabled"
+            "state": "enabled",
+            "minSupportedVersion": "0.22.3"
```
jonathanKingston, line 47: Was going to say I thought we still needed to strip this manually but the other code in index.js does this... thanks!

jonathanKingston (approved):

PR #227, opened 2022-05-04
Add custom UserAgent feature

SlayterDev:
https://app.asana.com/0/1200890834746050/1200932417493197/f


features/custom-user-agent.json
```diff
@@ -0,0 +1,19 @@
+{
+    "_meta": {
+        "description": "Custom user agents used by our native apps that override the default user-agent string.",
+        "sampleExcludeRecords": {
+            "domain": "example.com",
+            "reason": "Two factor auth that verifies device pathes using user agent"
+        }
+    },
+    "settings": {
+        "omitApplicationSites": [
```
jonathanKingston, line 10: Can't these be regular exceptions?

features/custom-user-agent.json
```diff
@@ -0,0 +1,19 @@
+{
+    "_meta": {
+        "description": "Custom user agents used by our native apps that override the default user-agent string.",
+        "sampleExcludeRecords": {
```
jonathanKingston, line 4: It's about time we drop this key IMO, it's not really relevant anymore.

features/custom-user-agent.json
```diff
@@ -0,0 +1,19 @@
...
+        }
+    },
+    "settings": {
+        "omitApplicationSites": [
+            "cvs.com",
+            "sovietgames.su",
+            "accounts.google.com",
+            "facebook.com"
+        ],
+        "omitVersionSites": []
```
jonathanKingston, line 16: What does this do?

SlayterDev: In Android they have two levels of exceptions for the UserAgent. 1 where they drop DuckDuckGo from the agent (omit application) and another where they drop the browser version (omit version). iOS does not drop the version, only the application name.

features/custom-user-agent.json
SlayterDev: Perhaps as a follow up or LHF? Feels weird to drop that from all features in this PR.

features/custom-user-agent.json
SlayterDev: They _could_. However given how they're used (selectively dropping parts of the custom UA) I don't think its appropriate. As mentioned below Android has a more granular approach. IMO exceptions on this feature should be used if _no_ UserAgent customization should happen at all.

features/custom-user-agent.json
jonathanKingston: I see, I personally was going to push for this kinda thing:

```
"exceptions": [
  { domain: "boop.com", "settings": {omitVersion: true}},
  { domain: "else.com" }
]
```

features/custom-user-agent.json
jonathanKingston: Yeah agreed

features/custom-user-agent.json
SlayterDev: This will significantly increase client side integration complexity as (at least on the Apple platforms) the exceptions structure is tightly defined within the config implementation.

jonathanKingston: This is something we're probably going to want for autofill where settings are going to be granular and we might want per site settings.
Creating distinct URL lists won't scale there overtime I don't think.

jonathanKingston: Also the way I was going to implement this in the extension would just be to merge settings with the domain, which makes the implementation less boilerplate and apply across all features.

SlayterDev: This seems like a config v3 discussion. IMO there's room for debate as tracker allowlist sets up precedence for keeping this sort of thing outside of exceptions.

aitorvs: > This will significantly increase client side integration complexity as (at least on the Apple platforms) the exceptions structure is tightly defined within the config implementation.

What do you think about fixing the Apple platforms rather than have them prevent what otherwise would be a better json structure like the one @jonathanKingston suggested?

jonathanKingston (changes requested):
As discussed.

features/custom-user-agent.json
kdzwinel: As discussed on the call - we are all aligned that `exceptions > {domain, settings}` seems like a better pattern moving forward, so let's get rid of the old one (`settings > bla: []`) across the whole config in v3 (we use same pattern in trackingCookies3p, trackerAllowlist, gpc).

overrides/ios-override.json
```diff
@@ -32,6 +32,9 @@
         },
         "autofill": {
             "state": "enabled"
+        },
+        "customUserAgent": {
+            "state": "enabled"
```
kdzwinel, line 37: 1. And @marcosholgado  will do the same for android during his project, right?

2. Is our android list of domains the same as iOS list? I'm guessing not, so should we move iOS domains into an overwrite or Marcos will do it?

3. to double-check -client implementation is not going into BSK and macOS is not affected, right?

kdzwinel (commented):
Looks good, I left some questions.

features/custom-user-agent.json
```diff
@@ -0,0 +1,19 @@
+{
+    "_meta": {
+        "description": "Custom user agents used by our native apps that override the default user-agent string.",
+        "sampleExcludeRecords": {
+            "domain": "example.com",
+            "reason": "Two factor auth that verifies device pathes using user agent"
```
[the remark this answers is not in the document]
SlayterDev: Will fix 👍

overrides/ios-override.json
SlayterDev: 1. Yes

2. I haven't seen their list so we'll move it as needed in their project.

3. Correct not going in to BSK at this time.

features/custom-user-agent.json
```diff
@@ -0,0 +1,19 @@
+{
+    "_meta": {
+        "description": "Custom user agents used by our native apps that override the default user-agent string.",
+        "sampleExcludeRecords": {
+            "domain": "example.com",
+            "reason": "Site Breakage"
```
kdzwinel, line 6:
```suggestion
            "reason": "Site doesn't recognize our user agent and blocks user from logging in."
```

sorry to spend so much time on this :) but I think a good example can be actually useful. I think google auth blocked us in the past so I tried to base the example on that experience

features/custom-user-agent.json
```diff
@@ -0,0 +1,19 @@
...
+        }
+    },
+    "settings": {
+        "omitApplicationSites": [
+            "cvs.com",
+            "sovietgames.su",
+            "accounts.google.com",
+            "facebook.com"
```
kdzwinel, line 14: hm, wait wait, I just realized that those don't match `sampleExcludeRecords`. We should have domain/reason here, right?

kdzwinel (approved):
thanks for those changes, LGTM

PR #228, opened 2022-05-05
Generate release body using etags (md5 hashes)

SlayterDev:
https://app.asana.com/0/1200890834746050/1202179697219704/f

This PR adds etags to the generated releases. Note this changes the action we're using for releases as the previous action didn't support setting the release body.


scripts/computeEtag.sh
```diff
@@ -0,0 +1,31 @@
...
+
+MD5_CMD=$1
+MD5_IDX=$2
+
+OUTPUTSTR=""
+
+for ver in generated/v*
+do
+    OUTPUTSTR+="**$(basename $ver)**\n"
+    OUTPUTSTR+="|Filename|etag|\n"
```
jonathanKingston, line 18: I think this whole file would be simpler in node.js personally; we're not using shell scripts anywhere else.

SlayterDev:
Proof of functionality can be seen here: https://github.com/SlayterDev/privacy-configuration/releases/tag/1657831715240

kdzwinel (approved):
Thanks for that preview. LGTM, neat improvement!

PR #234, opened 2022-05-18
Support non-tracking cookies and per browser configs

SlayterDev:
**Asana Task:** https://app.asana.com/0/72649045549333/1202156648532924/f
Also see: https://app.asana.com/0/1200890834746050/1202234294126144/f

**Description**
This PR adds the `non-tracking-3p-cookies` feature which is only used by Chrome. Which means this PR ALSO adds support for per browser configurations. The override dir now has a `browsers` subdir which contains override files for the user agents we support in the extension. If an unsupported browser requests the config it should fall back to `extension-config.json` (see https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1166).


index.js
```diff
@@ -98,17 +104,24 @@ const platformConfigs = {}
...
 for (const platform of platforms) {
-    const platformConfig = JSON.parse(JSON.stringify(defaultConfig))
+    let platformConfig = JSON.parse(JSON.stringify(defaultConfig))
     const overridePath = `${OVERRIDE_DIR}/${platform}-override.json`
 
+    // Use extension config as the base for browser configs
+    // Extension comes first in the list of platforms so its config should be defined
+    // in the platformConfigs array
+    if (platform.includes(BROWSERS_SUBDIR)) {
+        platformConfig = JSON.parse(JSON.stringify(platformConfigs.extension))
```
jonathanKingston, line 114: This relies on the sort order of ./platforms being exactly as it is.

Can you file a follow up to fix this and/or add a comment to the platforms.js to indicate ordering matters.

PR #240, opened 2022-05-27
Add unified cookies config

jonathanKingston:
**Asana Task:**

https://app.asana.com/0/0/1202348774384668/f

**Description**

Adds a unified cookie feature for: https://github.com/duckduckgo/content-scope-scripts/pull/78/

### DIFF
--- a/features/cookie.json
+++ b/features/cookie.json
@@ -1,0 +1,11 @@
+{
+    "_meta": {
+        "description": "Restriction of 1st party cookies and 3rd party cookie blocking",
+        "sampleExcludeRecords": {
+            "domain": "example.com",
+            "reason": "site breakage"
+        }
+    },
+    "exceptions": [
+    ]
+}
--- a/overrides/extension-override.json
+++ b/overrides/extension-override.json
@@ -3,6 +3,109 @@
         "autofill": {
             "state": "enabled"
         },
+        "cookie": {
+            "state": "enabled",
+            "settings": {
+                "trackerCookie": "enabled",
+                "nonTrackerCookie": "enabled",
+                "excludedCookieDomains": [
+                    {
+                        "domain": "hangouts.google.com",
+                        "reason": "Site breakage"
+                    },
+                    {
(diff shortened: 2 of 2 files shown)


features/cookie.json
```diff
@@ -0,0 +1,110 @@
+{
+    "_meta": {
+        "description": "Restriction of 1st party cookies and 3rd party cookie blocking",
+        "sampleExcludeRecords": {
+            "domain": "example.com",
+            "reason": "site breakage"
+        }
+    },
+    "state": "enabled",
```
SlayterDev, line 9: Do we want to enable this for all? Or just override it in Chrome? Likely this will need to be enabled on Chrome and FF and `nonTrackerCookie` disabled on FF.

jonathanKingston: "state" should be enabled for all and then the settings overriden in FF. I'll make the change; this was whilst testing :)

SlayterDev: Maybe just enabled for extensions? There's no use enabling it for native platforms and the config should be more clear in that regard.

overrides/browsers/chrome-override.json
```diff
@@ -1,3 +1,100 @@
 {
-    "features": { }
+    "features": {
+        "cookie": {
+            "settings": {
+                "trackerCookie": "enabled",
```
jonathanKingston, line 5: This duplication is unfortunate I raised: https://app.asana.com/0/1200890834746050/1202348774384704/f

overrides/extension-override.json
```diff
@@ -3,6 +3,109 @@
         "autofill": {
             "state": "enabled"
         },
+        "cookie": {
+            "state": "enabled",
+            "settings": {
+                "trackerCookie": "disabled",
```
SlayterDev, line 9: Think we'll want this enabled on extension. Only on Chrome are we doing `nonTrackerCookie`.

SlayterDev (approved):

SlayterDev:
🚀

PR #254, opened 2022-06-08
Use lodash deep merging to merge config overrides

SlayterDev:
**Asana Task:** https://app.asana.com/0/1200890834746050/1202348774384704/f

**Description**
This PR uses lodash's `mergewith` to merge overrides into platform configs. Added benefit is removing a large amount of merging code we were doing ourselves. I've added tests for some of these deep merging scenarios but they may not be exhaustive. Open to suggestions.

cc @jonathanKingston


index.js
```diff
@@ -64,6 +65,24 @@ function generateV1Config (platformConfig) {
...
+    }
+
+    if (Array.isArray(objValue)) {
+        return Array.from(new Set(objValue.concat(srcValue)))
+    }
+}
+
+function isFeatureMissingState (feature) {
+    return !('state' in feature)
+}
```
kdzwinel, line 84: when this can happen? Shouldn't that be caught by schema checking?

package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "test": "npm run lint && npm run unit-tests",
     "lint": "eslint .",
-    "unit-tests": "mocha ./tests/**/*.js"
+    "unit-tests": "mocha ./tests/**/*.js",
+    "clean": "rm -rf ./generated/*"
```
kdzwinel, line 10:
```suggestion
    "clean": "rm -rf ./generated/*.json"
```

rm -rf is kinda scary, let's limit to json if possible

tests/build-tests.js
```diff
@@ -1,174 +1,203 @@
...
-            {
-                rule: 'f3.com/foo',
-                domains: [
-                    'dom1.com',
-                    'dom0.com', // should be sorted in output
-                    'dom1.com' // duplicate should be removed
-                ],
-                reason: 'Reason1.'
+    features: {
+        trackerAllowlist: {
```
kdzwinel, line 70: why are we testing only trackerAllowlist? Is this a good representation of all possible features? e.g. it doesn't seem to have regular exceptions only settings/exceptions

tests/build-tests.js
```diff
@@ -1,174 +1,203 @@
...
+                        rules: [
+                            {
+                                rule: 'www.f1.com/foo/bar',
+                                domains: [
+                                    'dom0.com',
+                                    'dom1.com',
+                                    'dom2.net',
+                                    'dom3.net'
+                                ],
+                                reason: 'Reason1.; Reason2.'
```
kdzwinel, line 131: hm, interesting, where is the code that does this concatenation of reasons? `arrayMerge` doesn't seem to be doing this?

kdzwinel: Hm, I see some more tests in the config-tests file. Why those are separate?

kdzwinel (commented):
I left some minor comments, just trying to understand the code better.

package.json
SlayterDev: We also have some `.txt` files in this directory.

tests/build-tests.js
SlayterDev: There are other settings tests in config-tests. This was modified from Ben's code to test tracker allowlist merging but refactored to use the deep merge method added here.

index.js
SlayterDev: Schema checking tests after build. Since there are features that we exempt the default state (and explicitly enable in the overrides) this is just a safe-guard to ensure the key exists.

package.json
GuiltyDolphin: How about `rm -rf ./generated/*.{json,txt}`?

PR #288, opened 2022-07-04
Remove unnecessary cookie mitigations

GuiltyDolphin:
https://app.asana.com/0/0/1202535160524145/f

Removes mitigations that are no longer necessary e.g., because we can no longer reproduce breakage, or due to new privacy features superseding older ones.


kdzwinel (approved):
LGTM, nice job 🧹

PR #289, opened 2022-07-04
Update mitigation for foxnews

GuiltyDolphin:
https://app.asana.com/0/1200223097357040/1202121182005411/f

### DIFF
--- a/features/tracker-allowlist.json
+++ b/features/tracker-allowlist.json
@@ -266,14 +266,42 @@
                     }
                 ]
             },
-            "segment.io": {
+            "segment.com": {
                 "rules": [
+                    {
+                        "rule": "cdn.segment.com/analytics-next/bundles/130.bundle.487014ebc6f0bb448b6e.js",
+                        "domains": [
+                            "foxnews.com"
+                        ],
+                        "reason": "When on login page, clicking 'create account' has no effect."
+                    },
+                    {
+                        "rule": "cdn.segment.com/analytics-next/bundles/ajs-destination.bundle.5b76fa91c402ddfb4d2c.js",
+                        "domains": [
+                            "foxnews.com"
+                        ],
+                        "reason": "When on login page, clicking 'create account' has no effect."
+                    },
+                    {
+                        "rule": "cdn.segment.com/analytics-next/bundles/schemaFilter.bundle.a77eb8c5db3e65045afc.js",
+                        "domains": [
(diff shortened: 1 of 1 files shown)


features/tracker-allowlist.json
```diff
@@ -395,14 +395,42 @@
                     }
                 ]
             },
-            "segment.io": {
+            "segment.com": {
                 "rules": [
+                    {
+                        "rule": "cdn.segment.com/analytics-next/bundles/130.bundle.487014ebc6f0bb448b6e.js",
```
kdzwinel, line 401: this seems fragile - `/130.bundle.487014ebc6f0bb448b6e` - it suggests that they are regularly updating names of those libraries . Making rule a bit broader e.g. `cdn.segment.com/analytics-next/bundles/*.js` would reduce number of rules and be less fragile IMO.

GuiltyDolphin: @kdzwinel there are various requests that could be made that we don't want to match here - e.g., `segment.com/analytics-next/bundles/ajs-destination.bundle.a6950cf6bd0c8b0b0e97.js`. This should be fine as a temporary mitigation - we can review more general rules in investigations.

features/tracker-allowlist.json
```diff
@@ -395,14 +395,42 @@
...
                     {
                         "rule": "cdn.segment.com/analytics.js/v1/4I2fAtE5IxTlczNT3F91OckHtI5wfR7g/analytics.min.js",
                         "domains": [
                             "foxnews.com"
                         ],
-                        "reason": "user cannot access create account page from sign up page"
+                        "reason": "When on login page, clicking 'create account' has no effect."
+                    },
+                    {
+                        "rule": "cdn.segment.com/v1/projects/4I2fAtE5IxTlczNT3F91OckHtI5wfR7g/settings",
```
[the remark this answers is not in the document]
GuiltyDolphin: @kdzwinel the ID appears to be site-specific - let's only allow the specific request for the time being.

features/tracker-allowlist.json
kdzwinel: Gotcha, since this is temporary I'm OK with the precise rules.

kdzwinel (approved):
Not merging per the tag, but LGTM!

kdzwinel (approved):

PR #290, opened 2022-07-05
Remove unnecessary mitigations

GuiltyDolphin:
https://app.asana.com/0/0/1202552268487487/f

Removes mitigations that are no longer necessary e.g., because we can no longer reproduce breakage, or due to new privacy features superseding older ones.

### DIFF
--- a/features/autoconsent.json
+++ b/features/autoconsent.json
@@ -7,18 +7,6 @@
         }
     },
     "exceptions": [{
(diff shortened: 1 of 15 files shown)


kdzwinel (approved):
Nicely done 👏 As discussed - since you tested all of those and since great majority is single-domain-scoped I think it's fine to merge it all at once.

PR #306, opened 2022-07-12
Update mitigation reason fields with more information

GuiltyDolphin:
https://app.asana.com/0/0/1202556802923935/f

Updates the reason field for mitigations to provide more information about the breakage encountered.

This also adds support for splitting reason descriptions over multiple array elements, to make longer descriptions easier to read.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,6 @@
 const fs = require('fs')
 
-const { fixReasons, mergeAllowlistedTrackers } = require('./util')
+const { inlineReasonArrays, mergeAllowlistedTrackers } = require('./util')
 
 const OVERRIDE_DIR = 'overrides'
 const GENERATED_DIR = 'generated'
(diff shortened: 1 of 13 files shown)


index.js
```diff
@@ -27,8 +27,8 @@ function writeConfigToDisk (platform, config, v1Config) {
         configName = platform.replace(BROWSERS_SUBDIR, 'extension-')
     }
 
-    fs.writeFileSync(`${GENERATED_DIR}/v2/${configName}-config.json`, JSON.stringify(config, null, 4))
-    fs.writeFileSync(`${GENERATED_DIR}/v1/${configName}-config.json`, JSON.stringify(v1Config, null, 4))
+    fs.writeFileSync(`${GENERATED_DIR}/v2/${configName}-config.json`, JSON.stringify(fixReasons(config), null, 4))
+    fs.writeFileSync(`${GENERATED_DIR}/v1/${configName}-config.json`, JSON.stringify(fixReasons(v1Config), null, 4))
```
kdzwinel, line 31: I thought about it a bit more and I think we should change this code:

- having `fixReasons` in `writeConfigToDisk` is kinda unexpected (name suggests that we are only writing to disk), also, all other processing happens before this function. Let's move `fixReasons` to before `writeConfigToDisk` is called. If you call it before v1 config is generated you can probably get away with calling it once instead of twice per platform.
- I'd change the name from `fixRasons` to something more explicit e.g. `inlineReasonArrays`

kdzwinel (commented):
As discussed the updated reasons look good to me (nicely done!), but I have second thoughts about that inlining code - please see my comment.

PR #324, opened 2022-07-20
Add FP values to features and overrides

SlayterDev:
**Asana Task:** https://app.asana.com/0/72649045549333/1200810841420629/f

**Description**
This PR adds fingerprinting override values to the config and customizes them for macOS and iOS.

### DIFF
--- a/overrides/macos-override.json
+++ b/overrides/macos-override.json
@@ -36,9 +36,9 @@
         "fingerprintingHardware": {
             "state": "enabled",
             "settings": {
-                "keyboard": "undefined",
-                "hardwareConcurrency": 4,
-                "deviceMemory": "undefined"
+                "keyboard": {"type": "undefined"},
+                "hardwareConcurrency": {"type": "number", "value": 4},
+                "deviceMemory": {"type": "undefined"}
             }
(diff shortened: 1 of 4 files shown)


overrides/macos-override.json
```diff
@@ -33,6 +33,23 @@
             "state": "enabled",
             "minSupportedVersion": "0.22.3"
         },
+        "fingerprintingHardware": {
+            "state": "enabled",
+            "settings": {
+                "keyboard": {"type": "undefined"},
+                "hardwareConcurrency": [
+                    {"type": "number", "value": 4},
+                    {"type": "number", "value": 8, "criteria": {"arch": "AppleSilicon"}}
```
bwaresiak, line 42: Question: Is is a problem if we report more than actually available on a given device? I'd guess not.

SlayterDev: @bwaresiak I don't believe so. See deeper discussion here: https://app.asana.com/0/0/1202624285456073/f

bwaresiak (approved):
I've checked the tech design and discussion in asana, nothing stand out to me.

Issue #362, opened 2022-09-05
klaviyo.com breakage


GuiltyDolphin:
This issue outlines breakage for requests associated with the `klaviyo.com` domain.

- https://shopyalehome.com (last verified 2022-09-05)
  - site renders blank
  - requests required
    - `www.klaviyo.com/media/js/public/klaviyo_subscribe.js`
- https://fearofgod.com (last verified 2022-09-05)
  - site renders blank
  - requests required
    - `www.klaviyo.com/media/js/public/klaviyo_subscribe.js`

sammacbeth:
Further `klaviyo.com` breakage:

- https://manage.kmail-lists.com/subscriptions/
  - site renders blank
  - requests required: https://static-app.klaviyo.com/fender/83919.401de5abb7927d48beb1.js

Root cause is missing entity data - kmail-lists.com and klaviyo.com share the same owner. This has been fixed in tracker-radar (https://github.com/duckduckgo/tracker-radar/pull/168) and will be resolved in the next blocklist update.

chas-klaviyo:
Hi all, we have an additional issue that we opened up around this root issue: https://github.com/duckduckgo/privacy-configuration/issues/551 cc: @sammacbeth @GuiltyDolphin

jdorweiler:
- https://footweartruth.com/  email signup form is missing

shakyShane:
- email signup form missing from https://andieswim.com/en-gb/pages/fit-finder too

dharb:
Additional Klaviyo breakage:
- https://github.com/duckduckgo/privacy-configuration/issues/1277 - kidsguide.com
- https://github.com/duckduckgo/privacy-configuration/pull/1748 - muc-off.com, paria.cc

tagawa:
Email signup form missing on https://usafacts.org/signup/

kzar:
Email signup form missing on https://holdonbags.com/

kzar:
Email signup form missing on nestlenutritionstore.com

kzar:
Email signup form missing on https://dreamfarm.com

laghee:
Closing out as **archived** since we updated to a new transparency system in PR #2137

PR #363, opened 2022-09-06
Add klaviyo mitigation

GuiltyDolphin:
Mitigates some sites from #362


kdzwinel (approved):
This is probably a good case for a surrogate. Whole page is set to `opacity: 0` and it's just a matter of removing that one CSS rule. Mitigation LGTM!

PR #514, opened 2022-10-12
Add Duck Player feature

ayoy:
**Asana Task:**
https://app.asana.com/0/1200890834746050/1203149907948928/f

**Description**
This change adds Duck Player feature to macOS config. It's to be able to disable it remotely as needed.

### DIFF
--- a/features/duck-player.json
+++ b/features/duck-player.json
@@ -1,6 +1,6 @@
 {
     "_meta": {
-        "description": "Private YouTube video player",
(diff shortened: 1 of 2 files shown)


features/duck-player.json
```diff
@@ -0,0 +1,7 @@
+{
+    "_meta": {
+        "description": "Private YouTube video player",
```
SlayterDev, line 3: Could we expand this to be just a bit more descriptive? As this is public facing we're trying to keep these parts clear.

ayoy: Sure! How about now @SlayterDev?

SlayterDev (approved):
Thanks for this!

PR #521, opened 2022-10-18
Disable spiegel.de and derstandard.at

adewes:
**Asana Task:** https://app.asana.com/0/1201844467387842/1203191122685034/f & https://app.asana.com/0/1199230911884351/1203144940444659/f

**Description**

As we're still having issues correctly handling cookie consent on these sites (and as we can't do an opt-out there anyway) let's disable them for now until we have a fix.


GuiltyDolphin (approved):
We should switch these over to using issues, but I'll link to this PR for now.

PR #523, opened 2022-10-20
Disable autoconsent on swm.de to prevent infinite reload

sammacbeth:
**Asana Task:** https://app.asana.com/0/1201844467387842/1203197597853180/f

**Description**
False positive in popup open detection, and a page reload triggered by opt-out are causing an infinite reload on swm.de in the mac browser.


adewes (commented):
LGTM.

SlayterDev (approved):

PR #525, opened 2022-10-25
Disable autoconsent on asus.com due to broken opt-out

sammacbeth:
**Asana Task:** https://app.asana.com/0/1201844467387842/1203224632405176/f

**Description**


adewes:
LGTM. Any idea what's broken or how we can fix it long term?

sammacbeth:
@adewes on the site itself the code path when you opt out its an exception which stops it getting to the point where the overlay is hidden.

adewes (approved):
LGTM.

PR #531, opened 2022-10-31
Add 'Edg' platform for modern Edge browsers.

sammacbeth:
**Asana Task:**

**Description**
As discussed in https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1486, Chromium-based Edge will be resolved as 'Edg' by the extension. This PR adds a config for 'edg'.


jonathanKingston:
Should we remove the other one? It seems very unlikely to be used.

sammacbeth:
> Should we remove the other one? It seems very unlikely to be used.

My feeling is, I'd want to avoid it becoming a 404 for those browsers that might still be hitting it. As a code path still exists in the live version that explicitly calls that URL, it's not yet safe to remove it.

jonathanKingston (approved):

PR #532, opened 2022-10-31
Disable autoconsent on heise.de (infinite reload)

sammacbeth:
**Asana Task:** https://app.asana.com/0/1199178362774117/1203176209844998/f

**Description**
Also reported here https://github.com/duckduckgo/autoconsent/issues/35


adewes (approved):
Fine for me, do we have a strategy to fix this in the future (or an idea what exactly causes the behavior)?

PR #549, opened 2022-11-21
List all utm params

jdorweiler:
**Asana Task:**

**Description**
There are only 5 utm parameters https://en.wikipedia.org/wiki/UTM_parameters
We can list them out instead of using a regex.


SlayterDev:
This may require client side changes (at the very least requires discussion). We'll discuss it when I return from vacation 👍

SlayterDev (approved):

PR #585, opened 2022-12-16
Disable CPM on tfl.gov.uk

sammacbeth:
**Asana Task:**
https://app.asana.com/0/1200277586140538/1203520995494944/f

**Description**
https://github.com/duckduckgo/autoconsent/issues/68


muodov (approved):
This seems to be a race condition in autoconsent (sometimes it works correctly). We should remove this once we fix that bug.

PR #599, opened 2023-01-01
AppTP: Remove stale scoped exclusions

laghee:
**Asana Task:** https://app.asana.com/0/1202279501986195/1203633396255861/f

**Description**
We have many potentially stale tracker exceptions for various apps, none of which were being honored anyway until late November due to a bug in how we processed the list. I've removed all the exceptions older than November for which there is no obvious or compelling evidence (e.g., unblocking `www.instagram.com` to allow linking to Instagram within another app, particularly an app for which managing IG is its main function, would be considered a good reason).

Adding reasons for the exceptions that remain will be done in a subsequent PR.

### DIFF
--- a/features/app-tracker-protection.json
+++ b/features/app-tracker-protection.json
@@ -38,99 +38,10 @@
         "exceptionLists": {
             "appTrackerAllowList": [
                 {
-                    "domain": "acdn.adnxs.com",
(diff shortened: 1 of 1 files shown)


nshuba (approved):
LGTM - glad to get rid of these!

PR #634, opened 2023-01-19
Duplicate nintendo.com allowlist entry to nintendo.com.edgekey.net

sammacbeth:
Due to a bug in android exception matching

**Asana Task:**
https://app.asana.com/0/1200295047952574/1203770718751292/f

**Description**
The original allowlist entry does not work on android as the domain is first party.

### DIFF
--- a/features/tracker-allowlist.json
+++ b/features/tracker-allowlist.json
@@ -527,6 +527,13 @@
                             "aliexpress.us"
                         ],
                         "reason": "https://github.com/duckduckgo/privacy-configuration/issues/570"
+                    },
+                    {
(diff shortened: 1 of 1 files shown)


features/tracker-allowlist.json
```diff
@@ -1041,6 +1041,17 @@
                     }
                 ]
             },
+            "nintendo.com.edgekey.net": {
```
GuiltyDolphin, line 1044: Move this to be a rule in the existing `edgekey.net` entry.

GuiltyDolphin (changes requested):

GuiltyDolphin (approved):

PR #638, opened 2023-01-20
Disable element hiding on blick.ch

sammacbeth:
**Asana Task:**
https://app.asana.com/0/1200277586140538/1201956343511070/f

**Description**
Prevent hard adblocker wall.


dharb (approved):
Thanks! Will have to look into their detection method.

PR #643, opened 2023-01-25
Add serviceworkerBlocking feature

sammacbeth:
**Asana Task:**
https://app.asana.com/0/72649045549333/1202513251610479/f

**Description**
Adds a new feature to control blocking of service worker originated requests.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203821302367736

### DIFF
--- a/features/serviceworker-blocking.json
+++ b/features/serviceworker-blocking.json
@@ -1,0 +1,11 @@
+{
+    "_meta": {
+        "description": "Service worker blocking. Blocking of requests initiated by service workers",
+        "sampleExcludeRecords": {
+            "domain": "example.com",
+            "reason": "site breakage"
+        }
+    },
+    "state": "enabled",
+    "exceptions": []
+}
--- a/features/serviceworker-initiated-requests.json
+++ b/features/serviceworker-initiated-requests.json
@@ -1,11 +1,10 @@
 {
     "_meta": {
-        "description": "Service worker blocking. Blocking of requests initiated by service workers",
+        "description": "Blocking of requests initiated by service workers. Allows specific blocking exceptions for requests made from service workers",
         "sampleExcludeRecords": {
             "domain": "example.com",
             "reason": "site breakage"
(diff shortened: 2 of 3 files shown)


features/serviceworker-blocking.json
```diff
@@ -0,0 +1,11 @@
+{
+    "_meta": {
+        "description": "Service worker blocking. Blocking of requests initiated by service workers",
+        "sampleExcludeRecords": {
+            "domain": "example.com",
+            "reason": "site breakage"
+        }
+    },
+    "state": "enabled",
```
kzar, line 9: I don't think the state belongs in this file, but in the platform overrides (e.g. [extension-override.json](https://github.com/duckduckgo/privacy-configuration/blob/main/overrides/extension-override.json)). But thinking about it, perhaps we should only enable the feature for a more specific group first? I think we could enable it for only Firefox extension users for example.

features/serviceworker-blocking.json
```diff
@@ -0,0 +1,11 @@
+{
```
kzar, line 1: I mean up to you, but perhaps `serviceworker-initiated-requests.json` would be a better name? Most of the other feature names seem to be the thing they're blocking or protecting against. Also, I figure "ServiceWorker blocking" might be a kind of confusing name, making folks think we are looking to block ServiceWorkers entirely?

features/serviceworker-blocking.json
```diff
@@ -0,0 +1,11 @@
...
+    "_meta": {
+        "description": "Service worker blocking. Blocking of requests initiated by service workers",
```
kzar, line 3: Maybe ditch the first sentence and expand the rest a bit? E.g. "Blocking of tracking requests initiated by service workers. Considered a separate feature since it has caused breakage in the past."

kzar (approved):
LGTM

PR #650, opened 2023-01-30
Add subfeatures to Android autofill config

CDRussell:
**Asana Task:**
https://app.asana.com/0/0/1203855914488234/f

**Description**
Adds finer-grained control over `autofill` by adding subfeatures, each of which can be independently enabled/disabled. 

All default to `disabled` currently as the feature is not yet generally available.


overrides/android-override.json
```diff
@@ -27,7 +27,20 @@
             "state": "enabled"
         },
         "autofill": {
-            "state": "enabled"
+            "state": "enabled",
+            "settings": {
+                "features": {
+                    "inject_credentials": {
```
marcosholgado, line 33: I believe we use camelcase rather than snake case, @SlayterDev can confirm.

CDRussell: Yeah I see that elsewhere now; updated

SlayterDev (approved):

PR #664, opened 2023-02-07
Added the exceptions that fix the site breakages caused by 3rd party cookie blocker on Windows

RendijsSmukulis:
**Asana Task:**
https://app.asana.com/0/72649045549333/1200810841420637/f

**Description**
- Ensures the "cookie" feature is lit up only starting with v0.36.0 release (i.e. upcoming release)
- Adds the cookie name/domain exceptions to reduce breakage in login flows in certain sites (or using certain auth providers). See https://app.asana.com/0/0/1203869087349414/f for more information on the broken sites and the fixes.
- The excludedCookieDomains are inherited from android overrides.

### DIFF
--- a/overrides/windows-override.json
+++ b/overrides/windows-override.json
@@ -11,6 +11,80 @@
         },
         "elementHiding": {
             "state": "enabled"
+        },
+        "cookie": {
+            "exceptions": [],
+            "state": "enabled",
+            "minSupportedVersion": "0.36.0",
+            "settings": {
+                "trackerCookie": "enabled",
+                "nonTrackerCookie": "disabled",
+                "excludedCookieDomains": [
+                    {
+                        "domain": "accounts.google.com",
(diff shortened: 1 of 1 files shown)


overrides/windows-override.json
```diff
@@ -11,6 +11,80 @@
         },
         "elementHiding": {
             "state": "enabled"
+        },
+        "cookie": {
+            "exceptions": [],
+            "state": "enabled",
+            "minSupportedVersion": "0.36.0",
+            "settings": {
+                "trackerCookie": "enabled",
```
szanto90balazs, line 20: Is "trackerCookie" and "nonTrackerCookie" a windows only attribute? If so I would suggest slightly renaming it to "trackerCookieBlocking", "nonTrackerCookieBlocking"

SlayterDev: This (technically) applies to all products, but mostly applies to the browser extensions. Their names should not change at this time.

SlayterDev (approved):

PR #660, opened 2023-02-06
Move autofill features one level up

aitorvs:
**Asana Task:**
https://app.asana.com/0/0/1203898052213010/f

**Description**
Move `autofill` `features` JSON object one level up, instead of being
inside `settings`

Making this change in overrides for now but in the future it is expected
that all sub-features follow this pattern

Also setting `additionalProperties` to `true` to avoid test from failing

PR #714, opened 2023-02-27
Add internally enabled autofill subfeatures to ios override

graeme:
Asana Task:
https://app.asana.com/0/481882893211075/1203938569856702/f

Description

- Adds finer-grained control over autofill by adding subfeatures, each of which can be independently enabled/disabled.
  - Similar to equivalent Android PR #650
- Utilises new structure discussed in linked Asana task
  - Similar to equivalent Android PR #660
- Toggle set to `internal` to represent [current iOS feature flag state](https://github.com/duckduckgo/iOS/blob/17ffefb0e48cabe5243ea6577011e9f99f091067/Core/FeatureFlagger.swift#L47)

I'd like someone from iOS as well as Privacy to check this over before merging.

### DIFF
--- a/overrides/ios-override.json
+++ b/overrides/ios-override.json
@@ -40,7 +40,18 @@
             "state": "enabled"
         },
         "autofill": {
-            "state": "enabled"
+            "state": "enabled",
+            "features": {
+                "injectCredentials": {
+                    "state": "disabled"
+                },
+                "saveCredentials": {
+                    "state": "disabled"
+                },
+                "accessCredentialManagement": {
+                    "state": "disabled"
+                }
+            }
         },
         "referrer": {
             "state": "enabled"


overrides/ios-override.json
```diff
@@ -40,7 +40,18 @@
             "state": "enabled"
         },
         "autofill": {
-            "state": "enabled"
+            "state": "internal",
```
THISISDINOSAUR, line 43: If my current understanding of how the existing autofill flag is used, this will be a problem. I think setting the top level flag to internal will break email for existing users

graeme: The reason I did this was that I couldn't see anywhere that's using the Privacy Config's `.autofill` key. We only seem to be using the `FeatureFlagger`.

SlayterDev: @graeme the trouble here is the content scope scripts. The config in its entirety is passed to the content scope scripts which execute JavaScript in every tab. These scripts are using the `autofill` key to control the autofill feature on the webpages.

graeme: Oh right yes, of course. I was told about this.

SlayterDev (dismissed):

SlayterDev:
All good from me. I'll let @THISISDINOSAUR have a final say if she has anything.

graeme:
On further discussion with Elle, I think the chosen flags here needs a bit more thought. I'm going to spike this a bit more to double check if these are the right / enough flags

graeme:
@SlayterDev After the discussions in the tech-design and some testing, I think this is good to go. In the nature of the maintainer docs comment "We strive to make the most minimally disrupting change possible." I removed the `emailProtection` key for now as it's not necessary for the more time-dependent autofill work and can be added later.

PR #721, opened 2023-03-02
Add "clickToLoad" feature configuration and deprecate "clickToPlay"

kzar:
The Click to Load feature was misnamed "clickToPlay" in the extension
configuration, which had a knock-on effect with the naming of files in
the content-scope-scripts and other repositories. Let's fix that now,
or at least start the towards fixing that.

We can't remove the old "clickToPlay" feature configuration yet, but
we can deprecate it and add the new "clickToLoad" feature
configuration ready for use. Also, let's take care to remove the old
"clicksBeforeSimpleVersion" setting from the new configuration and
correct the Facebook entity name ("Facebook, Inc." instead of
"Facebook") to match our block list.

### DIFF
--- a/features/click-to-load.json
+++ b/features/click-to-load.json
@@ -1,6 +1,6 @@
 {
     "_meta": {
-        "description": "DEPRECATED - See clickToLoad.",
+        "description": "Click to Load feature for blocking embedded third-party social/video content.",
         "sampleExcludeRecords": {
             "domain": "example.com",
             "reason": "site breakage"
@@ -8,9 +8,11 @@
     },
     "exceptions": [],
     "settings": {
-        "Facebook": {
-            "clicksBeforeSimpleVersion": 3,
+        "Facebook, Inc.": {
             "ruleActions": ["block-ctl-fb"]
+        },
+        "Youtube": {
+            "ruleActions": ["block-ctl-yt"]
         }
     }
 }
--- a/index.js
+++ b/index.js
@@ -166,7 +166,7 @@ async function buildPlatforms () {
                             settings[settingsKey] = overrideSettings[settingsKey]
                         }
                         platformConfig.features[key][platformKey] = settings
-                    } else if (key === 'clickToPlay' && platformKey === 'settings') {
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -184,12 +184,14 @@ async function buildPlatforms () {
             }
 
             // Ensure the correct enabled state for Click to Load entities.
-            const clickToLoadSettings = platformConfig?.features?.clickToPlay?.settings
```
kzar, line 187: FWIW I think the lack of check for `key === 'clickToPlay'` here was a bug. We were updating the disabled/enabled status for Click to Load entities again for every feature.

features/click-to-load.json
```diff
@@ -0,0 +1,18 @@
...
+        "description": "Click to Load feature for blocking embedded third-party social/video content.",
+        "sampleExcludeRecords": {
+            "domain": "example.com",
+            "reason": "site breakage"
+        }
+    },
+    "exceptions": [],
+    "settings": {
+        "Facebook, Inc.": {
+            "ruleActions": ["block-ctl-fb"]
```
kdzwinel, line 12: we don't need `clicksBeforeSimpleVersion` setting, from the click-to-play config, anymore?

overrides/extension-override.json
```diff
@@ -46,6 +46,14 @@
         "referrer": {
             "state": "enabled"
         },
+        "clickToLoad": {
```
kdzwinel, line 49: old version of the extension would ignore unknown keys, right?

kdzwinel (commented):
Left some questions, but other than that LGTM.

overrides/extension-override.json
kzar: Yea, I think so. I used the config editor to try using the new config that included "clickToLoad" and it seemed fine.

features/click-to-load.json
kzar: Nope, that's deprecated.

index.js
kdzwinel: FWIW we talked about impact of this bug with @kzar  and this wasn't a bug that was causing trouble, just being unoptimal and it only started causing trouble when Dave added [key] logic. Anyway, we are all good here.

PR #777, opened 2023-03-29
Element hiding: catch some edge case sign in with google popups

dharb:
**Asana Task:** https://app.asana.com/0/1203557590179903/1204283501453247/f

**Description**
This PR catches a couple variants of the "Sign in with Google" popup that weren't already covered. Also addresses some new ad spaces on investing.com.

cc @jonathanKingston @GuiltyDolphin

### DIFF
--- a/features/element-hiding.json
+++ b/features/element-hiding.json
@@ -204,6 +204,10 @@
                 "type": "hide"
             },
             {
+                "selector": "#credentials-picker-container",
+                "type": "hide"
+            },
+            {
                 "selector": "#credential_picker_iframe",
                 "type": "hide"
             },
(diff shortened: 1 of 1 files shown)


features/element-hiding.json
```diff
@@ -894,12 +928,29 @@
...
+            },
             {
                 "domain": "tripadvisor.com",
                 "rules": [
                     {
                         "selector": "#tripGoogleOnetapContainer",
                         "type": "hide"
+                    },
+                    {
+                        "selector": ".mxEhR",
```
dharb, line 952: This looks like it's a generated class that might change frequently, but I checked The Wayback Machine and it's been the same since at least December.

GuiltyDolphin (approved):

Issue #794, opened 2023-03-31
Additional Breakage: Experimental


GuiltyDolphin:
This task outlines select breakage that we cannot currently reproduce directly, but for which we have suspected breakage (e.g., from user reports).

This currently includes (from user reports):
- capitalone.com (https://app.asana.com/0/0/1204285623263283/f)
- navyfederal.org (https://app.asana.com/0/0/1204285623263283/f)
- att.com (https://app.asana.com/0/0/1204285623263283/f)
- wellsfargo.com (https://app.asana.com/0/0/1204285623263283/f)
- suncoastcreditunion.com (https://app.asana.com/0/0/1204348393104084/f)
- coles.com.au (2023-05-11) (https://app.asana.com/0/1200277586140538/1204570089939667/f)
  - we've received various reports about this site loading blank on Firefox
- fidelity.com (2023-05-29)
  - https://app.asana.com/0/1200277586140538/1204693762946202/f
  - reports of being unable to use various aspects of the site, broken logins, pages not loading, etc.
- citi.com (2023-06-16)
  - https://app.asana.com/0/1200289808319188/1204839880944305/f
  - reports of login flow not completing
  - also citibankonline.com (https://app.asana.com/0/1200277586140538/1205209113013098/f)
- ticketmaster.com (2023-06-19)
  - bot detection wall
  - see also https://github.com/duckduckgo/privacy-configuration/issues/575
  - Asana: https://app.asana.com/0/1200277586140538/1204824365461091/f
- foxnews.com (2023-06-22)
  - sustained reports with unknown root cause
  - https://app.asana.com/0/1200277586140538/1204683037000654/f
- usps.com (2023-06-23)
  - reports of login failure
  - this is likely due to canvas fingerprinting protection, indirectly confirmed
  - https://app.asana.com/0/1200277586140538/1204876346444061/f
- bank.marksandspencer.com (2023-06-23)
  - reports of issues when trying to load credit card statements
  - https://app.asana.com/0/1200277586140538/1204895350597254/f
- iscorp.com (2023-08-03)
  - reports of issues immediately after sign-in with unknown root cause
  - https://app.asana.com/0/1200277586140538/1205189838780921/f
- synchrony.com, mysynchrony.com, synchronybank.com (2023-08-04)
  - reports of login breakage, interaction with Amazon
  - https://app.asana.com/0/1200277586140538/1205209227479636/f

[...]

PR #795, opened 2023-03-31
Add exceptions for canvas fp and runtime checks on several sites with reported login breakage

dharb:
**Asana Task:** https://app.asana.com/0/0/1204285623263283/f

Mitigations for https://github.com/duckduckgo/privacy-configuration/issues/794

**Description**
We've seen an uptick in reported login breakage on several popular domains over the past week. We've been able to repro the issue on a few sites where we have accounts and traced it back to anti canvas fingerprinting, so we are disabling it on these sites until we understand what's going on. Out of an abundance of caution, we are also disabling runtime checks on these same sites.

cc @jonathanKingston

### DIFF
--- a/features/fingerprinting-canvas.json
+++ b/features/fingerprinting-canvas.json
@@ -23,12 +23,20 @@
             "reason": "https://github.com/duckduckgo/privacy-configuration/issues/499"
         },
         {
+            "domain": "att.com",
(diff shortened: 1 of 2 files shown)


GuiltyDolphin (approved):
Thanks, LGTM :)

PR #835, opened 2023-04-13
Enable duck player on windows via remote config

q71114:
**Asana Task:**
https://app.asana.com/0/0/1204389062547165/f

**Description**
This is needed for the upcoming Duck Player feature in Windows. 
Existing Windows app versions won't be affected as currently there's no support for Duck Player.


shakyShane:
@q71114 can we either:

- pause on merging this, because I have some 'settings' that we should merge at the same time
- or, merge and then we can follow up with an additional PR tomorrow/monday

CC: @jonathanKingston

shakyShane:
@q71114  @gkerenyi I'm talking this through with @jonathanKingston as we speak - looking to find a preferred format ASAP :

jonathanKingston (approved):
Looks good, thanks for making the changes to the format :)

PR #843, opened 2023-04-17
Github automated generated code output

jonathanKingston:
**Asana Task:**

https://app.asana.com/0/1200890834746050/1204208396328588/f

**Description**

Creates an automation that responds with generated file outputs to reduce the risk of incorrect authoring and review.

### DIFF
--- a/.github/scripts/diff-directories.js
+++ b/.github/scripts/diff-directories.js
@@ -24,7 +24,7 @@ function readFilesRecursively (directory) {
 }
 
 /**
- * Removes the version from the file contents to imrpove diff readability
+ * Removes superfluous info from the file contents to improve diff readability
  * @param {string} fileContent
  * @param {string} filePath
  * @returns {string}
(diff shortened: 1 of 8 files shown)


.github/scripts/diff-directories.js
```diff
@@ -0,0 +1,159 @@
+const fs = require('fs')
+const path = require('path')
+const diff = require('diff')
+
+function readFilesRecursively (directory) {
+    const filenames = fs.readdirSync(directory)
+    const files = {}
```
SlayterDev, line 7: Any fears that if the config or platforms grow that this object will become unmanageably large?

jonathanKingston: I wouldn't say so, this should manage 1000s of files without it being a performance boon I would say. (we could always cut off the legacy ones if becomes an issue too)

Were there other concerns about number of files?

SlayterDev: Mostly just memory related concerns. We'll see if it becomes an issue later on ✨

PR #855, opened 2023-04-20
Start CODEOWNERS file

jonathanKingston:
**Asana Task:**

https://app.asana.com/0/1200890834746050/1204111016484104/f

**Description**

Initial commit of some CODEOWNERS isn't complete at all, we can add more as we go.

I'll continue in the issue with onboarding the other reviewers after this lands.


SlayterDev (changes requested):
One minor change

SlayterDev (approved):

PR #866, opened 2023-04-24
Add showtime.com the eme exceptions

marcosholgado:
**Asana Task:**
https://app.asana.com/0/414730916066338/1204454663335476/f
**Description**
This PR adds an eme exception for showtime.com

### DIFF
--- a/overrides/android-override.json
+++ b/overrides/android-override.json
@@ -20,7 +20,7 @@
                 },
                 {
                     "domain": "showtime.com",
-                    "reason": "Site shows a blank page, preventing users to access it."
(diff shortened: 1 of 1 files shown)


SlayterDev:
@marcosholgado could you file an issue in this repo following the template with details about the issue? Then link the issue in the reason field of the exception please.

marcosholgado:
> @marcosholgado could you file an issue in this repo following the template with details about the issue? Then link the issue in the reason field of the exception please.

Hey @SlayterDev I'm not sure that makes sense for this specific case. This is a website that unless we allow eme permissions it won't work and we don't that by default only via exceptions. Do you mean adding a link to the Asana task where this was raised?

Issue #911, opened 2023-05-10
Click to Load YouTube breakage on Google search results page


kzar:
### Background
The "Click to Load" YouTube feature blocks embedded third-party YouTube videos, while giving the user the option to click to load the video. Unfortunately, the feature does not yet work correctly for embedded videos on the Google search results page, causing videos and video previews to break intermittently. As a result, we are disabling the feature on Google search domains.

### What to change
- Add domain exceptions for the "Click to Load" feature, so that it is not active for the Google search results pages. [See the list of those domains](https://www.google.com/supported_domains).

PR #912, opened 2023-05-10
Add "Click to Load" exceptions for Google search domains.

kzar:
The "Click to Load" YouTube feature does not work correctly for inline
YouTube videos viewed on the Google search results page, causing
videos and video previews to break intermittently. As a result, we are
disabling the feature on Google search domains.

See https://github.com/duckduckgo/privacy-configuration/issues/911


kzar:
@jdorweiler I've just tested this in a development build of the extension, seems fine. I notice the integration tests are failing, but that seems to be due to the generated diff being too large to display.

jonathanKingston:
We really should strip 'reason' from the output given that no client uses it. It's a complete waste of bytes over the wire tbh.

jdorweiler (approved):

PR #918, opened 2023-05-11
enable styletag injection

dharb:
**Asana Task:** https://app.asana.com/0/0/1204574150177460/f

**Description**
We previously disabled this feature because it was triggering CSP reports on sites with very strict CSPs. However, we've received a few complaints that this is causing us to not hide 'sign in with Google' banners on sites that load them after a delay, so it's clear we need to re-enable this.

After a bit of research, it looks like the fix we made in https://github.com/duckduckgo/content-scope-scripts/pull/314 to avoid triggering CSP reports in Firefox actually ended up triggering more CSP reports on other platforms - injecting a `<link rel="stylesheet">` tag on non-Firefox platforms violates CSPs with `style-src 'unsafe-inline'`, which is pretty common.

This is resolved in https://github.com/duckduckgo/content-scope-scripts/pull/508.


features/element-hiding.json
```diff
@@ -332,7 +332,12 @@
                 "type": "hide-empty"
             }
         ],
-        "styleTagExceptions": [],
+        "styleTagExceptions": [
+            {
```
jonathanKingston, line 336: Do you want to add ddg also, they trigger CSP errors for this too.

dharb:
Let's wait for https://github.com/duckduckgo/content-scope-scripts/pull/508 before merging this.

GuiltyDolphin (approved):

PR #1034, opened 2023-06-21
mitigate focus.de

icodebyamanda:
**Asana Task/Github Issue:** https://app.asana.com/0/0/1204862417002333/f

## Description
Adding a selector to element-hiding to mitigate focus.de blank space

### DIFF
--- a/features/element-hiding.json
+++ b/features/element-hiding.json
@@ -172,6 +172,10 @@
                 "type": "hide-empty"
             },
             {
+                "selector": "#top-banner",
+                "type": "hide"
+            },
+            {
                 "selector": "#topBanner",
                 "type": "hide-empty"
             },
--- a/features/element-hiding.json
+++ b/features/element-hiding.json
@@ -566,24 +566,23 @@
                     {
                         "selector": "[id*='M_CONTENTAD']",
                         "type": "closest-empty"
-                      },
-                      {
+                    },
+                    {
                         "selector": "[id*='M_TRSCT_hor']",
                         "type": "closest-empty"
-                      },
-                      {
+                    },
+                    {
                         "selector": "[id*='M_FLYCPT_hor']",
                         "type": "closest-empty"
-                      },
-                      {
+                    },
+                    {
                         "selector": "#tfm_admanagerTeaser",
                         "type": "closest-empty"
-                      },
-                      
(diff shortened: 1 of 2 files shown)


features/element-hiding.json
```diff
@@ -171,6 +171,10 @@
                 "selector": ".top-banner",
                 "type": "hide-empty"
             },
+            {
+                "selector": "#top-banner",
+                "type": "hide"
```
GuiltyDolphin, line 176: Do we want this to be `hide-empty` so we don't start hiding populated content?

icodebyamanda: yes - change made

features/tracker-allowlist.json
```diff
@@ -1117,6 +1117,14 @@
                     }
                 ]
             },
+                    },
```
GuiltyDolphin, line 1120: Uh the formatting sems weird here. Did something go awry during a rebase/merge?

There should be an existing entry for `greylabeldelivery.com`, so you can add there.

FYI, the "reason" field should link to a GitHub issue (descriptive comments are deprecated).

features/element-hiding.json
```diff
@@ -334,7 +338,27 @@
...
+                "type": "closest-empty"
+              },
+              {
+                "selector": "[id*='M_FLYCPT_hor']",
+                "type": "closest-empty"
+              },
+              {
+                "selector": ".cls_slot_xxoutstreamxx",
+                "type": "hide-empty"
+              }
```
dharb, line 361: These are specific to focus.de, yeah? Let's move them out of the global section so they aren't applied on all pages and instead add them under the `domains` section, eg:
```JSON
{
    "domain": "focus.de",
    "rules": [
        {
            "selector": "[id*='M_CONTENTAD']",
            "type": "closest-empty"
        }
    ]
}
```

features/tracker-allowlist.json
dharb: What Ben said, see https://github.com/duckduckgo/privacy-configuration/pull/949 for an example of a content blocking mitigation.

dharb (changes requested):
Left a couple comments

dharb (approved):
Looks great, thank you!

Issue #1066, opened 2023-06-30
cdnjs.cloudflare.com breakage


GuiltyDolphin:
- https://kartor.eniro.se/ (2023-06-30)
  - map (i.e., full page) doesn't render
  - https://app.asana.com/0/1200277586140538/1204928062074300/f
  - requests required:
    - `cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.7/leaflet.js`
- 2000mules.com (2023-06-30)
  - Loading percentage hits 100% and the site does not proceed to display usual content.
  - requests required:
    - `cdnjs.cloudflare.com/ajax/libs/three.js/84/three.min.js`
- https://www.winnipegfreepress.com/local/city-will-try-to-fill-money-losing-ex-canada-post-office-576646662.html
  - Main text content on article does not load
  - requests required:
    - `cdnjs.cloudflare.com/ajax/libs/fingerprintjs2/1.8.6/fingerprint2.min.js`
      - (note: 2023-06-30, presently unable to get this request to fire)

laghee:
Closing out as **archived** since we updated to a new transparency system in PR #2137

PR #1067, opened 2023-06-30
Mitigate eniro maps

GuiltyDolphin:
Mitigates eniro maps in #1066.

https://app.asana.com/0/1200277586140538/1204928062074300/f

### DIFF
--- a/features/tracker-allowlist.json
+++ b/features/tracker-allowlist.json
@@ -412,6 +412,13 @@
             "cloudflare.com": {
                 "rules": [
                     {
+                        "rule": "cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.7/leaflet.js",
+                        "domains": [
+                            "eniro.se"
+                        ],
+                        "reason": "https://github.com/duckduckgo/privacy-configuration/issues/1066"
+                    },
+                    {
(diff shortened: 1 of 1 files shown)


sammacbeth (approved):

features/tracker-allowlist.json
```diff
@@ -409,22 +409,26 @@
             },
             "cloudflare.com": {
                 "rules": [
+                    {
+                        "rule": "cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.7/leaflet.js",
```
dharb, line 413: @GuiltyDolphin we should expand this rule to not block leaflet.js anywhere, as it's not a tracker and it's a common map component.

Looking at [other sites](https://raw.githubusercontent.com/duckduckgo/tracker-radar/main/domains/US/cloudflare.com.json) that fetch leaflet from the same endpoint, I see quite a bit of breakage (eg https://map.krak.dk/).

I think we should consider unblocking three.js everywhere on line 427 below as well, as it's a popular js library used for graphics and not a tracker.

GuiltyDolphin: @dharb could you add some sites you've repro'd the breakage on to the issue, then we can generalise the mitigation

PR #1151, opened 2023-07-25
Add config v3

SlayterDev:
**Asana Task/Github Issue:** https://app.asana.com/0/0/1205112086862872/f

## Description
This PR adds v3 of the config which will support incremental feature rollouts. It also generalizes backwards compatibility to make future versions easier to manage.

#### Reference


SlayterDev:
**Note:** We'll need to wait til this is merged to update the diff script as there is no reference v3 directory to diff.

index.js
```diff
@@ -39,18 +40,37 @@ async function getTds () {
...
     let configName = platform
     if (platform.includes(BROWSERS_SUBDIR)) {
         configName = platform.replace(BROWSERS_SUBDIR, 'extension-')
     }
 
-    // Now that the features have been finalized, write out their hashes
+    // Convert config to backwards compatible versions
+    let compatConfig = config
+    const configsToWrite = []
+    for (const version of ['v2', 'v1']) {
```
jonathanKingston, line 52: Would this be easier to maintain to reduce issues as for loop that counts down to the minimal version?

The functions could then be an array of functions? Or to retain the same structure append the v to the integer.

index.js
```diff
@@ -39,18 +40,37 @@ async function getTds () {
...
+        if (!compatibility.compatFunctions[version]) {
+            continue
```
jonathanKingston, line 54: This seems unexpected, should we throw here to prevent broken config parsing and uploading?

index.js
```diff
@@ -119,6 +117,7 @@ mkdirIfNeeded(GENERATED_DIR)
 // Create version directories
 mkdirIfNeeded(`${GENERATED_DIR}/v1`)
 mkdirIfNeeded(`${GENERATED_DIR}/v2`)
+mkdirIfNeeded(`${GENERATED_DIR}/v3`)
```
jonathanKingston, line 120: This probably should be driven from the same data structure if possible. So enumerate the array/object and construct all these directories.

index.js
```diff
@@ -39,18 +40,37 @@ async function getTds () {
...
+        fs.writeFileSync(`${GENERATED_DIR}/${version}/${configName}-config.json`, JSON.stringify(config, null, 4))
+    }
+
     addHashToFeatures(config)
-    addHashToFeatures(v1Config)
 
-    fs.writeFileSync(`${GENERATED_DIR}/v2/${configName}-config.json`, JSON.stringify(config, null, 4))
-    fs.writeFileSync(`${GENERATED_DIR}/v1/${configName}-config.json`, JSON.stringify(v1Config, null, 4))
+    // Write current version to disk
+    fs.writeFileSync(`${GENERATED_DIR}/v3/${configName}-config.json`, JSON.stringify(config, null, 4))
```
jonathanKingston, line 73: Should this come out of the compatiblity module? or have something keep track of the current version so we can construct the directories from it and iterate backwards from n?

compatibility.js
```diff
@@ -0,0 +1,65 @@
+function versionToInt (version) {
+    // convert v2 to 2
+    return parseInt(version.replace('v', ''))
+}
+
+function removeEolFeatures (config, prevConfig, version) {
+    for (const feature of Object.keys(config.features)) {
+        const eol = config.features[feature].eol
+        if (eol && versionToInt(eol) < version) {
```
jonathanKingston, line 9: version is an int here already in it's usages

compatibility.js
```diff
@@ -0,0 +1,65 @@
...
+            // This feature's support ends in a previous config so remove it from platformConfig
+            delete config.features[feature]
+
+            if (versionToInt(eol) === version - 1) {
+                // Only remove the eol key if it is the previous version
+                delete prevConfig.features[feature].eol
```
jonathanKingston, line 15: How can you delete something that's already deleted above?

compatibility.js
```diff
@@ -0,0 +1,65 @@
+function versionToInt (version) {
+    // convert v2 to 2
+    return parseInt(version.replace('v', ''))
+}
+
+function removeEolFeatures (config, prevConfig, version) {
```
jonathanKingston, line 6: This function is very hard to follow. At least document it more but I think it should be rethought to prevent issues.

My understanding is you're passing in `config` which is the current live config and removing features based on the current compat function.

- These side effects of editing two object references solve the problem but it feels unrelated to the current config you're operating on.

I would suggest it would be much simpler to follow to have a separate step ran after the `compatFunctions` to remove the relevant attributes/features from that config version only.

- That means that each call site `compatFunction` wouldn't have to wire in the call to this method potentially passing the wrong arguments.
- I would place this call just before the call to `addHashToFeatures(compatConfig)`
- Instead after running `compatFunction` `v2` you would delete anything that was EOL'd in v1
- This step also needs to apply to the current version so it doesn't really belong in the compat function.

compatibility.js
```diff
@@ -0,0 +1,65 @@
...
+        // Disable sub-features with a rollouts key.
+        // This key is not supported in v2 and any sub-features with this key
+        // will need to be disabled.
+        for (const feature of Object.keys(v2Config.features)) {
+            const subFeatures = v2Config.features[feature].features
+            if (subFeatures) {
+                for (const subFeature of Object.keys(subFeatures)) {
+                    if (subFeatures[subFeature].rollouts) {
+                        delete subFeatures[subFeature].rollouts
+                        v2Config.features[feature].features[subFeature].state = 'disabled'
```
jonathanKingston, line 51: So many features need to be re-written to match this subfeatures schema, it's a little disappointing tbh.

compatibility.js
SlayterDev: I toyed around with this approach and it's probably more correct. I ended up here after trying to generalize the v1 function. I'll rework it this way though as it does seem more clean.

index.js
jonathanKingston: Specifically I was talking about the `v3`.

Thinking about this some more, it would be nice if it was all part of the same for loop, starting with the latest version and counting down then on successive loops calling the relevant compat methods and writing the file.

compatibility.js
jonathanKingston: I'll be around tomorrow to help/explain further in case what I wrote makes no sense as it's late here haha 💤

jonathanKingston (approved):

PR #1297, opened 2023-09-08
Move global unprotected temporary exceptions to selected features on build

laghee:
**Asana Task/Github Issue:** https://app.asana.com/0/1200890834746050/1202346295795542/f

## Description
We shouldn't be globally excluding certain non-explicitly-privacy-protecting or otherwise-unrelated-to-web-pages features when we mitigate for breakage. This change distributes the domain exceptions individually to only the features we should disable in these cases.

#### Reference

### DIFF
--- a/index.js
+++ b/index.js
@@ -104,14 +104,12 @@ const listData = JSON.parse(fs.readFileSync(`${LISTS_DIR}/${unprotectedListName}
 addExceptionsToUnprotected(listData.exceptions)
 addExceptionsToUnprotected(defaultConfig.features.contentBlocking.exceptions)
 
-const notExcluded = ['adClickAttribution', 'appTrackerProtection', 'autofill', 'duckPlayer', 'incontextSignup',
+const excludedFeaturesFromUnprotectedTempExceptions = ['adClickAttribution', 'appTrackerProtection', 'autofill', 'duckPlayer', 'incontextSignup',
     'incrementalRolloutTest', 'networkProtection', 'newTabContinueSetUp', 'voiceSearch', 'windowsPermissionUsage',
     'windowsWaitlist', 'windowsDownloadLink']
-for (const exception of listData.exceptions) {
-    for (const key of Object.keys(defaultConfig.features)) {
-        if (!(notExcluded.includes(key))) {
-            defaultConfig.features[key].exceptions = defaultConfig.features[key].exceptions.concat(exception)
-        }
+for (const key of Object.keys(defaultConfig.features)) {
+    if (!(excludedFeaturesFromUnprotectedTempExceptions.includes(key))) {
+        defaultConfig.features[key].exceptions = defaultConfig.features[key].exceptions.concat(listData.exceptions)
     }
 }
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -100,11 +101,20 @@ function addExceptionsToUnprotected (exceptions) {
 }
 
 const listData = JSON.parse(fs.readFileSync(`${LISTS_DIR}/${unprotectedListName}`))
-defaultConfig.unprotectedTemporary = listData.exceptions
-
-addExceptionsToUnprotected(defaultConfig.unprotectedTemporary)
+addExceptionsToUnprotected(listData.exceptions)
 addExceptionsToUnprotected(defaultConfig.features.contentBlocking.exceptions)
 
+const notExcluded = ['adClickAttribution', 'appTrackerProtection', 'autofill', 'duckPlayer', 'incontextSignup',
```
jonathanKingston, line 107: `notExcluded` could be read either way as not excluded from this treatment or not an exclusion.
Could we consider using a name that adds a little more clarity like `excludedFeaturesFromUnprotectedTemporaryExceptions` (maybe there's something less wordy it's late here ha)

index.js
```diff
@@ -100,11 +101,20 @@ function addExceptionsToUnprotected (exceptions) {
...
+    'incrementalRolloutTest', 'networkProtection', 'newTabContinueSetUp', 'voiceSearch', 'windowsPermissionUsage',
+    'windowsWaitlist', 'windowsDownloadLink']
+for (const exception of listData.exceptions) {
+    for (const key of Object.keys(defaultConfig.features)) {
+        if (!(notExcluded.includes(key))) {
+            defaultConfig.features[key].exceptions = defaultConfig.features[key].exceptions.concat(exception)
```
jonathanKingston, line 113: `concat` can accept an array, we can remove the `const exception of listData.exceptions` and make this line:
```suggestion
            defaultConfig.features[key].exceptions = defaultConfig.features[key].exceptions.concat(listData.exceptions)
```

index.js
laghee: Didn't have any kind of brainstorm on shortening that aside from `Temp` for `Temporary` assuming we want to keep all the context 🤷‍♀️

index.js
```diff
@@ -100,11 +101,18 @@ function addExceptionsToUnprotected (exceptions) {
 }
 
 const listData = JSON.parse(fs.readFileSync(`${LISTS_DIR}/${unprotectedListName}`))
-defaultConfig.unprotectedTemporary = listData.exceptions
-
-addExceptionsToUnprotected(defaultConfig.unprotectedTemporary)
+addExceptionsToUnprotected(listData.exceptions)
 addExceptionsToUnprotected(defaultConfig.features.contentBlocking.exceptions)
 
+const excludedFeaturesFromUnprotectedTempExceptions = ['adClickAttribution', 'appTrackerProtection', 'autofill', 'duckPlayer', 'incontextSignup',
```
SlayterDev, line 107: This is a nit but could we break this list out to one entry per line for readability?

SlayterDev, line 107: Let's maybe also drop a comment above this line to explain what the list is.

SlayterDev (commented):
Thank you @laghee! I've verified the generated output looks correct. I've added a couple nits here just to improve the readability.

SlayterDev (approved):
LGTM!

PR #1325, opened 2023-09-26
Strip reasons from v4+

SlayterDev:
**Asana Task/Github Issue:** https://app.asana.com/0/1200890834746050/1202306580815233/f

## Description
This PR updates the build script to remove reasons from the generated config files from version 4 onwards. Reasons will still be available in the source files and will point to open GitHub issues in [this repo](https://github.com/duckduckgo/privacy-configuration/issues).

#### Reference

### DIFF
--- a/compatibility.js
+++ b/compatibility.js
@@ -39,49 +39,71 @@ function removeEolFeatures (config, version) {
  * return {object} - The converted config object
  */
 const compatFunctions = {
-    v1: {
-        stripReasons: false,
-        compat: (config) => {
-            // Breaking changes: minSupportedVersion key in features
+    v1: (unmodifiedConfig, config) => {
+        // Breaking changes: minSupportedVersion key in features
 
-            const v1Config = JSON.parse(JSON.stringify(config))
-            for (const feature of Object.keys(v1Config.features)) {
-                if (v1Config.features[feature].minSupportedVersion) {
-                    delete v1Config.features[feature].minSupportedVersion
-                    v1Config.features[feature].state = 'disabled'
-                }
+        const v1Config = JSON.parse(JSON.stringify(config))
+        for (const feature of Object.keys(v1Config.features)) {
+            if (v1Config.features[feature].minSupportedVersion) {
+                delete v1Config.features[feature].minSupportedVersion
+                v1Config.features[feature].state = 'disabled'
             }
-
-            return v1Config
         }
-    },
-    v2: {
-        stripReasons: false,
-        compat: (config) => {
-            // Breaking changes: rollout key in sub-features
 
-            const v2Config = JSON.parse(JSON.stringify(config))
-            for (const feature of Object.keys(v2Config.features)) {
-                const subFeatures = v2Config.features[feature].features
-                if (subFeatures) {
-                    for (const subFeature of Object.keys(subFeatures)) {
-                        if (subFeatures[subFeature].rollout) {
-                            delete subFeatures[subFeature].rollout
-                            v2Config.features[feature].features[subFeature].state = 'disabled'
-                        }
+        return v1Config
+    },
+    v2: (unmodifiedConfig, config) => {
+        // Breaking changes: rollout key in sub-features
+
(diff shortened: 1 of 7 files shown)


compatibility.js
```diff
@@ -39,36 +39,49 @@ function removeEolFeatures (config, version) {
...
             }
-        }
 
-        return v2Config
+            return v2Config
+        }
+    },
+    v3: {
+        stripReasons: false,
+        compat: (config) => {
```
jonathanKingston, line 81: It might be simpler to pass in the unmodified config as the first argument to these functions and on v3 merge back in reasons from the unmodified config.

That way if we ever have the need to do this kinda thing in future we'd not have to pass more arguments.

tests/config-tests.js
```diff
@@ -12,17 +12,18 @@ function formatErrors (errors) {
     return errors.map(item => `${item.instancePath}: ${item.message}`).join(', ')
 }
 
+// Test the latest 2 versions of each platform
 const v2configs = platforms.map((plat) => {
```
jonathanKingston, line 16: These variable names don't match what we're testing. Should we rename or test all variants?

tests/schemas/exception.json
```diff
@@ -5,5 +5,5 @@
         "reason": {"type": "string"}
     },
     "additionalProperties": false,
-    "required": ["domain", "reason"]
```
jonathanKingston, line 8: I think this is a problem, unless we're going to add a blank reason code also as new exceptions will be added and Android will expect the property.

compatibility.js
```diff
@@ -39,7 +39,7 @@ function removeEolFeatures (config, version) {
  * return {object} - The converted config object
  */
 const compatFunctions = {
-    v1: (config) => {
+    v1: (unmodifiedConfig, config) => {
```
jonathanKingston, line 42: I'd consider swapping these, config is the primary functionality here and `unmodifedConfig` should be used less. This would allow v1, v2 to omit the argument too.

compatibility.js
```diff
@@ -69,6 +69,41 @@ const compatFunctions = {
...
+        for (const key of Object.keys(unmodifiedConfig.features)) {
+            for (const i in unmodifiedConfig.features[key].exceptions) {
+                v3Config.features[key].exceptions[i].reason = unmodifiedConfig.features[key].exceptions[i].reason
+            }
+
+            if (key === 'trackerAllowlist') {
+                for (const domain of Object.keys(unmodifiedConfig.features[key].settings.allowlistedTrackers)) {
+                    for (const i in unmodifiedConfig.features[key].settings.allowlistedTrackers[domain].rules) {
+                        v3Config.features[key].settings.allowlistedTrackers[domain].rules[i].reason = unmodifiedConfig.features[key].settings.allowlistedTrackers[domain].rules[i].reason
+                    }
```
jonathanKingston, line 88: Untested but you could reduce the level of boilerplate like this:

```suggestion
                    const rules = v3Config.features[key].settings.allowlistedTrackers[domain].rules
                    rules = rules.map((rule, i) => {
                        const reason = unmodifiedConfig.features[key].settings.allowlistedTrackers[domain].rules[i].reason
                        return Object.assign(rule, { reason })
                    })
```
This reduces the risk of `i`'s and `domains` and whatever being in the wrong order.

The latter part could be a shared function between the other rules potentially too.

jonathanKingston (commented):
Some comments attached :)

compatibility.js
jonathanKingston: Also we could consider making sure reason is present here too.

```js
                    const rules = v3Config.features[key].settings.allowlistedTrackers[domain].rules
                    rules = rules.map((rule, i) => {
                        const reason = unmodifiedConfig?.features?.[key]?.settings?.allowlistedTrackers?.[domain]?.rules?.[i].reason || ""
                        return Object.assign(rule, { reason })
                    })
```

bwaresiak (approved):

PR #1376, opened 2023-10-10
Element hiding: temporarily disable styletag injection

dharb:
## Description
Temporarily disabling styletag injection until the bugfix in https://github.com/duckduckgo/content-scope-scripts/pull/775 is released on all platforms. Strict hide rules will still apply as expected, the only noticeable difference is it might take a second for them to be applied (instead of instantly). Will re-enable just as soon as clients update to C-S-S 4.39.0.

#### Reference

### DIFF
--- a/features/element-hiding.json
+++ b/features/element-hiding.json
@@ -85,7 +85,7 @@
         }
     ],
     "settings": {
-        "useStrictHideStyleTag": true,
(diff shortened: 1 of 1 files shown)


jonathanKingston (approved):
Thanks for fixing this! Glad we added this option :)

PR #1414, opened 2023-10-19
Remove unnecessary element hiding exceptions

kzar:
Some websites attempt to detect ad blockers by checking if an element
with a common class (e.g. class="ad") is hidden. Sometimes those
websites will display a "Please disable your ad blocker" message to
users when they suspect an ad blocker is active. To avoid getting
caught up in that, we added some websites to our element hiding
exception list. That prevented us from hiding any elements for those
websites.

Let's go back through the list and remove which ones we can. Some of
those ad walls were actually fixed by our "surrogate" script
improvements[1][2][3]. Where necessary though, let's use the new
"disable-default" element hiding rule (which works similarly to
$generichide[4]) so that we avoid hiding elements with the usual
selectors. While we're at it, let's also add the relevant element
hiding rules to hide any blank spaces left by blocking tracking
scripts (since they often relate to ads) that still aren't hidden on
those websites.

1 - https://github.com/duckduckgo/tracker-surrogates/commit/ec20dfba47bd5ceef76216c083c8317c1c1691a3
2 - https://github.com/duckduckgo/tracker-surrogates/commit/75f4e54124d72b820728ea8d25727a87e2293f26
3 - https://github.com/duckduckgo/tracker-surrogates/commit/acf0c0fd89502af696a112126178511b84f21e72
4 - https://blog.adblockplus.org/development-builds/new-filter-options-generichide-and-genericblock

**Asana Task/Github Issue:** https://app.asana.com/0/0/1205594040522080/f

### DIFF
--- a/features/element-hiding.json
+++ b/features/element-hiding.json
@@ -10,78 +10,6 @@
         {
             "domain": "duckduckgo.com",
             "reason": "https://github.com/duckduckgo/privacy-configuration/issues/1236"
-        },
-        {
-            "domain": "bild.de",
-            "reason": "https://github.com/duckduckgo/privacy-configuration/issues/589"
-        },
-        {
-            "domain": "derstandard.at",
-            "reason": "https://github.com/duckduckgo/privacy-configuration/issues/1101"
-        },
-        {
-            "domain": "foxnews.com",
-            "reason": "https://github.com/duckduckgo/privacy-configuration/issues/965"
-        },
-        {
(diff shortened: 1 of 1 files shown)


dharb:
Thanks @kzar, looks great! But let's wait to merge this until the new surrogates are shipped on all clients so we don't inadvertently retrigger adblock walls on clients that haven't updated yet.

dharb (approved):
👍

PR #1477, opened 2023-11-08
New variantManager remote config parameter

nalcalag:
**Asana Task/Github Issue:** https://app.asana.com/0/0/1205908353497982/f

## Description
New top level parameter `variantManager` has been added to the Android privacy configuration in order to manage experiments variants remotely. 
- SERP variants will remain there with a _weight 0.0_, ensuring they are not assigned to users unless we decide to change it.
- I also added `zj` and `zh` variants because we currently have an experiment running, and we need to keep them in the remote config to avoid their removal from existing users until the [project](https://app.asana.com/0/72649045549333/1205122649889482/f) is finished

#### Reference

### DIFF
--- a/overrides/android-override.json
+++ b/overrides/android-override.json
@@ -431,7 +431,7 @@
     },
     "unprotectedTemporary": [
     ],
-    "variantManager": {
+    "experimentalVariants": {
         "variants": [
           {
             "desc": "this is SERP don't remove",
--- a/overrides/android-override.json
+++ b/overrides/android-override.json
@@ -446,12 +446,18 @@
           {
             "desc": "Control variant for AskForDefaultBrowserMoreThanOnce experiment",
             "variantKey": "zh",
-            "weight": 0.0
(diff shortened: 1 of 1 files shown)


overrides/android-override.json
```diff
@@ -645,5 +645,35 @@
         }
     },
     "unprotectedTemporary": [
-    ]
+    ],
+    "variantManager": {
```
SlayterDev, line 649: Let's make the naming here more clear. Something like `experimentVariantManager` will make it more clear what this section is for.

```suggestion
    "experimentVariantManager": {
```

SlayterDev (changes requested):
Thank you for this PR @nalcalag. Just one material comment to make this feature a bit more transparent.

overrides/android-override.json
nalcalag: @SlayterDev I changed the name for `experimentalVariants` as we agreed on the Asana task 👍

PR #1662, opened 2023-12-20
Sync feature - transfer subfeatures and global state to feature json

cmonfortep:
**Asana Task/Github Issue:** https://app.asana.com/0/0/1206190482102399/f

## Description
Decided to have a global config instead of a per platform override.

#### Reference

### DIFF
--- a/features/sync.json
+++ b/features/sync.json
@@ -3,7 +3,7 @@
         "description": "Sync & Backup browser feature. Feature available on native clients. Allow users to sync bookmarks, logins, etc.",
         "sampleExcludeRecords": {}
     },
-    "state": "internal",
+    "state": "disabled",
     "features": {
         "level0ShowSync": {
             "state": "enabled"
--- a/features/sync.json
+++ b/features/sync.json
@@ -3,5 +3,20 @@
         "description": "Sync & Backup browser feature. Feature available on native clients. Allow users to sync bookmarks, logins, etc.",
         "sampleExcludeRecords": {}
     },
+    "state": "internal",
+    "features": {
+        "level0ShowSync": {
+            "state": "enabled"
+        },
+        "level1AllowDataSyncing": {
+            "state": "enabled"
+        },
+        "level2AllowSetupFlows": {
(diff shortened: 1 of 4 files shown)


ayoy (commented):
@cmonfortep could you please update iOS and macOS overrides to remove sync completely? That's because their current overrides are going to the top level definition with this PR. Thanks!

cmonfortep:
oh I didn't know you already added them! you are so fast @ayoy

ayoy:
It's actually @SlayterDev who is doing a wonderful job reviewing all the PRs within minutes 💪🏻💪🏻💪🏻

ayoy:
@cmonfortep sorry to be a pain, but there's also the macos-override.json that needs to be updated. I'm AFK so can't do it myself at the moment. Thanks 😇

features/sync.json
```diff
@@ -3,5 +3,20 @@
         "description": "Sync & Backup browser feature. Feature available on native clients. Allow users to sync bookmarks, logins, etc.",
         "sampleExcludeRecords": {}
     },
+    "state": "internal",
```
SlayterDev, line 6: Hate to be a pain but we should probably default disable this on all platforms and just set it to internal where needed. We don't want any unintended side effects on other platforms even if they don't use the feature.

SlayterDev (changes requested):

features/sync.json
bwaresiak: I wonder about that. Is 'internal' a state that is expected to be globally supported? If not, are there plans for that? Cause what I'd expect it to be able to use internal state exactly in situations like this - where we are developing something and we have it as an internal feature across products (for products that don't support given feature it should not matter - as they won't parse the flag in the first place).

cmonfortep: In theory it should work like @bwaresiak said, but I also prefer to err on the side of being careful here and leave default state as `disabled`.  At least is consistent with the value when the state is not present.

SlayterDev (approved):

PR #2128, opened 2024-06-07
Introduce new feature flag for Improve Mobile Site Breakage Report Quality project

jaceklyp:
**Asana Task/Github Issue:** https://app.asana.com/0/72649045549333/1207379730878832/f

### DIFF
--- a/overrides/ios-override.json
+++ b/overrides/ios-override.json
@@ -417,6 +417,9 @@
                     "state": "enabled"
                 }
             }
+        },
+        "brokenSiteReportExperiment": {
+            "state": "enabled"
         }
     },
     "unprotectedTemporary": []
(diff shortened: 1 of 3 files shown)


overrides/ios-override.json
```diff
@@ -417,6 +417,9 @@
                     "state": "enabled"
                 }
             }
+        },
+        "brokenSiteReportExperiment": {
+            "state": "enabled"
```
bwaresiak, line 422: I believe this is supposed to be temporary feature, so we can remove it once experiment is done. 

@SlayterDev is there a playbook for adding features? do we have a pattern to handle temp pixels? Shall we prefix the name with something to make it clear? Is there a list of used names so we don't by accident  "resurrect" some experiment on some older versions? :)

SlayterDev: Our playbook for new features is here: https://app.asana.com/0/1200890834746050/1201498956177210/f

I don't think we have any specifics around temp features. I would lean on the side of prefixing with `temp*` just for clarity sake if we intend to remove them. And we currently don't maintain a list of old names. I'm not sure if we've ever actually removed a feature entirely :)

bwaresiak (dismissed):

PR #2187, opened 2024-07-04
Enable deduplication of imported logins on macOS

graeme:
**Asana Task/Github Issue:**
https://app.asana.com/0/0/1207675780406461/f

## Description

We're adding a feature to remove duplicate autofill login entries from imported data. We should ship with this on but, in the case that we receive negative user feedback from the results, we should also provide a quick way to disable it remotely.

This adds the autofill sub-feature and sets it to be enabled (it's not been shipped yet so this is low risk).

#### Reference

### DIFF
--- a/overrides/macos-override.json
+++ b/overrides/macos-override.json
@@ -487,7 +487,12 @@
             "state": "enabled"
         },
         "autofill": {
-            "state": "enabled"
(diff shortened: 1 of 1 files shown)


federicocappelli (approved):
LGTM

PR #2203, opened 2024-07-16
iOS & Android: Adds new autofillBreakageReporter feature flag

amddg44:
**Asana Task/Github Issue:**
https://app.asana.com/0/1201645688934642/1207323756661954/f
cc: @CDRussell 

## Description

This change introduces the `autofillBreakageReporter` feature to be used on both mobile platforms (iOS & Android) to dynamically control access to a new UI feature that allows users to report autofill breakages. The feature also includes the option for site exceptions and the setting `monitorIntervalDays` which will allow us to dynamically control how many days it will be before a user can report the same site again.

#### Reference

### DIFF
--- a/overrides/android-override.json
+++ b/overrides/android-override.json
@@ -230,6 +230,14 @@
                 }
             }
         },
+        "autofillBreakageReporter": {
+            "state": "enabled",
+            "settings": {
+                "monitorIntervalDays": 42
+            },
+            "exceptions": []
+        },
+
         "androidBrowserConfig": {
             "state": "enabled",
             "features": {
--- a/overrides/android-override.json
+++ b/overrides/android-override.json
@@ -231,11 +231,7 @@
             }
         },
         "autofillBreakageReporter": {
-            "state": "enabled",
-            "settings": {
-                "monitorIntervalDays": 42
-            },
-            "exceptions": []
+            "state": "enabled"
         },
         "androidBrowserConfig": {
             "state": "enabled",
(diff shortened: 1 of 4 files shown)


SlayterDev (changes requested):
Please create a base file for this feature in the `features` directory. You can add the settings block to that file since they appear to be the same, then override the state to `enabled` in the two override files.

overrides/android-override.json
```diff
@@ -230,6 +230,13 @@
                 }
             }
         },
+        "autofillBreakageReporter": {
+            "state": "enabled",
+            "settings": {
```
SlayterDev, line 235: This block can be removed since its already declared in the base file.

SlayterDev, line 235: This block can be removed since its already declared in the base file.

overrides/ios-override.json
```diff
@@ -150,6 +150,13 @@
                 }
             }
         },
+        "autofillBreakageReporter": {
+            "state": "enabled",
+            "settings": {
```
SlayterDev, line 155: Same here.

SlayterDev (changes requested):
Just a bit of clean up :)

SlayterDev (changes requested):
Sorry @amddg44 one more thing I should've mentioned earlier. Please add your feature to this list: https://github.com/duckduckgo/privacy-configuration/blob/d7b8c10728896057aeea8b9e01aed5752580d976/index.js#L116

SlayterDev (approved):
Perfect! Thank you for all those changes.

PR #2318, opened 2024-09-23
Add a feature flag for the bottom omnibar

0nko:
**Asana Task/Github Issue:** https://app.asana.com/0/1207418217763355/1208319036523676/f

## Description

- Adds a feature flag for the bottom address bar feature on Android.
- With two toggles:
   - One to disable the whole bottom bar feature, including the settings option (the default value will be `internal`).
   - One to disable the refactored omnibar component specifically and use the legacy omnibar.

#### Reference

### DIFF
--- a/overrides/android-override.json
+++ b/overrides/android-override.json
@@ -1271,6 +1271,14 @@
         },
         "androidNewStateKillSwitch": {
             "state": "disabled"
+        },
+        "changeOmnibarPosition": {
+            "state": "internal",
+            "features": {
+                "refactor": {
+                    "state": "disabled"
+                }
+            }
         }
     },
     "unprotectedTemporary": [],
@@ -1313,13 +1321,5 @@
                 }
             }
         ]
-    },
-    "changeOmnibarPosition": {
-        "state": "internal",
-        "features": {
-            "refactor": {
-                "state": "disabled"
-            }
-        }
     }
 }
--- a/overrides/android-override.json
+++ b/overrides/android-override.json
@@ -1313,5 +1313,13 @@
                 }
             }
         ]
+    },
+    "changeOmnibarPosition": {
+        "state": "internal",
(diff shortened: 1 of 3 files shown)


overrides/android-override.json
```diff
@@ -1313,5 +1313,13 @@
...
             }
         ]
+    },
+    "changeOmnibarPosition": {
+        "state": "internal",
+        "features": {
+            "refactor": {
+                "state": "disabled"
+            }
+        }
```
aitorvs, line 1323: @0nko, where is the `changeOmnibarPosition` and `refactor` flags used in the Android code. We already have an `androidBrowserConfig` feature, maybe we just add a sub-feature in threre?

0nko: @aitorvs It’s used [here](https://github.com/duckduckgo/Android/pull/4885/files#diff-afc0d0350ec07d2ca8049fc263053a34faf4841263abc688f8c65294ede7be68). The `refactor` feature will be used by @malmstein in a subsequent refactoring of the omnibar. I don’t have any preference here, David suggested to make these changes so I’ll let him decide what’s better.

aitorvs (dismissed):

overrides/android-override.json
```diff
@@ -1313,5 +1313,13 @@
                 }
             }
         ]
+    },
+    "changeOmnibarPosition": {
```
SlayterDev, line 1317: @0nko This block must be inside the `features` block of the override file. Currently it is out side of everything so the feature still appears disabled in the android config.

SlayterDev (changes requested):

overrides/android-override.json
0nko: Oops, thank you! Fixed.

SlayterDev (approved):
LGTM!

PR #2353, opened 2024-10-15
Add Signal to AppTP unprotected list.

greyson-signal:
## Description
Hi there, we've had some users report that they're unable to use [Signal](https://play.google.com/store/apps/details?id=org.thoughtcrime.securesms&hl=en_US) properly while AppTP is enabled. There's been reports of people having issues sending/receiving messages. I personally cannot reproduce the issue, but these particular users were adamant that the issue is tied to AppTP and goes away when they disable it. One of the users has a Moto G Power (pretty sure it's [this one](https://www.gsmarena.com/motorola_moto_g_power-10076.php)?) running Android 11.

So I made this PR to add us to the exclusion list by default, but I'll leave it up to ya'll on whether that's the best course of action, or whether there's more information I can get you.

### DIFF
--- a/features/app-tracker-protection.json
+++ b/features/app-tracker-protection.json
@@ -200,14 +200,6 @@
                     ]
                 },
                 {
-                    "domain": "www.facebook.com",
-                    "packageNames": [
-                        {
-                            "packageName": "org.thoughtcrime.securesms"
-                        }
-                    ]
-                },
-                {
                     "domain": "www.googletagmanager.com",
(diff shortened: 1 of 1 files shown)


aitorvs (approved):
LGTM

greyson-signal:
Yep! We'll attempt to reach out to the user to see if the problem has been resolved. Fingers crossed they respond :sweat_smile:  Thank you!

PR #2439, opened 2024-11-08
Use Typescript for config schemas

sammacbeth:
**Asana Task/Github Issue:** https://app.asana.com/0/481882893211075/1208516641945153/f

## Description
 - Defines the full config schema in TS.
 - Updates tests to validate v4 and v3 generated configs against this schema.
 - Replaces all existing JSONSchema definitions with TS equivalents, and fixes missing values where validation found them. Updates those tests to use the TS validator.

#### Reference


jonathanKingston:
Before we merge this I want to make sure C-S-S doesn't explode. We had some auto routing of the types so the `getSettings` code just magically worked with these settings types.

No real issues otherwise from a quick glance.

schema/config.d.ts
```diff
@@ -0,0 +1,61 @@
...
+    | 'WebCompatSettings'
+    | 'DuckPlayerSettings';
+
+/**
+ * Defines the structure of the built V4 config output as downloaded by clients.
+ */
+export type ConfigV4<VersionType> = {
+    readme: string;
+    version: number;
+    features: Record<string, Feature<any, VersionType>> & {
```
jonathanKingston, line 25: The previous one, checked for feature name matching a format. Can you raise a follow up issue to make sure we add a unit test? (Unless there's a way to do this in TS)?

sammacbeth: Does https://github.com/duckduckgo/privacy-configuration/pull/2439/commits/4e6f8f620699c08b2190301bb817c40961bf9844 cover it?

jonathanKingston: Thanks!

schema/config.d.ts
```diff
@@ -0,0 +1,61 @@
...
+        // These features have typed settings
+        autoconsent: AutoconsentFeature<VersionType>;
+        cookie: CookieFeature<VersionType>;
+        duckPlayer: DuckPlayerFeature<VersionType>;
+        trackerAllowlist: TrackerAllowlistFeature<VersionType>;
+        webCompat: WebCompatFeature<VersionType>;
```
jonathanKingston, line 31: Just so I understand: Does the webcompat-test.js add anything else type wise?  (Same for duckplayer)

jonathanKingston (approved):
LGTM, I played around with it a bit and also linked it into C-S-S fairly easily (no other consumers exist to these types that I'm aware of).

There's a bunch of fast follows like:

1. Typing out some of the C-S-S stuff to be shared a bit better to help feature authors with the domain patching stuff but that's certainly not a blocker as we didn't have anything here.
2. I think we can remove both duckplayer and webcompat test now this change links the testing in automatically for those settings. (Happy to remove these as a follow up if that's the case).

On risk to merge:

1. It looks like you've copied over what typing we already had (and added more).
Given that until recently we weren't even blocking merges on tests passing; we can take the risk on some edge cases of these being a little looser than they might have been (I couldn't spot any other than the feature name that you fixed :) ).
2. Additionally if the typing is slightly too restrictive, I think it's not a huge concern either.

Feel free to merge when ready 👍🏻

schema/config.d.ts
sammacbeth: No - we're just applying a subset of the schema on a subset of the data. I just kept the existing tests there, but the full schema test will be testing the same thing now, so these are redundant.

PR #2482, opened 2024-11-15
Fix pageload issue on itsthevibe.com

kzar:
It turns out that some of the pages on
itsthevibe.com (e.g. https://itsthevibe.com/most-loyal-dog-breeds/) fail to load
when these requests are blocked. It seems that a value that they set on the
"FCCDCF" property of some Object is later read back, if that fails the page
stops loading. Let's unblock those for now, to get them working again.

Note: There are also some rendering issues in the DuckDuckGo iOS browser, but
      those are also present in Safari.

**Asana Task/Github Issue:** https://app.asana.com/0/1200277586140538/1208762277986372/f

#### Reference

### DIFF
--- a/features/tracker-allowlist.json
+++ b/features/tracker-allowlist.json
@@ -790,6 +790,7 @@
                         "domains": [
                             "ah.nl",
                             "applesfera.com",
+                            "itsthevibe.com",
                             "nytimes.com",
                             "wunderground.com",
                             "youmath.it",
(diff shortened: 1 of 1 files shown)


tagawa (changes requested):
I think the reason should be the PR URL (https://github.com/duckduckgo/privacy-configuration/pull/2482) rather than descriptive text (which you have well covered in the PR description).

tagawa (approved):
Thanks for the update.

PR #2502, opened 2024-11-22
Add Malicious Site Protections Configuration

not-a-rootkit:
**Asana Task/Github Issue:** https://app.asana.com/0/72649045549333/1208815326657419/f

## Description
Currently the malicious site protections config is named phishingDetection but as we're adding more data classes to it I'm proposing we rename it to maliciousSiteProtection for consistency.
The config currently contains the following fields:
 - allowErrorPage: when true, the phishing detection feature is enabled and checks are performed for all navigation events.
 - allowPreferencesToggle: when true, there is a settings toggle to allow users to enable/disable the feature.

We've discussed removing the subfeature flags for `allowErrorPage` and `allowPreferencesToggle` so it's combined in just one main flag for the whole feature, and adding settings:

 -  when the main feature flag is enabled, we allow phishing + malware protections to run in the browser
 - settings.hashPrefixUpdateFrequency: currently set to 20 minutes
 - settings.filterSetUpdateFrequency: currently set to 12 hours (in minutes: 720)

### Feature change process:

- [x] I have added a [schema](https://github.com/duckduckgo/privacy-configuration/tree/main/schema) to validate this feature change
- [x] I have tested this change locally
- [x] This code for the config change is ready to merge - yes for Android + macOS.
- [x] This feature was covered by a tech design: https://app.asana.com/0/481882893211075/1207273224076495/f

#### Additional info:

#### Reference

### DIFF
--- a/features/malicious-site-protection.json
+++ b/features/malicious-site-protection.json
@@ -1,6 +1,6 @@
 {
     "_meta": {
-        "description": "Phishing Detection",
+        "description": "Malicious Site Protection",
         "sampleExcludeRecords": {}
     },
     "state": "disabled",
--- a/index.js
+++ b/index.js
@@ -141,6 +141,7 @@ const featuresToIncludeTempUnprotectedExceptions = [
     'navigatorInterface',
     'nonTracking3pCookies',
     'performanceMetrics',
+    'phishingDetection',
     'maliciousSiteProtection',
     'referrer',
     'requestFilterer',
--- a/index.js
+++ b/index.js
@@ -141,8 +141,6 @@ const featuresToIncludeTempUnprotectedExceptions = [
     'navigatorInterface',
     'nonTracking3pCookies',
     'performanceMetrics',
-    'phishingDetection',
-    'maliciousSiteProtection',
     'referrer',
     'requestFilterer',
     'runtimeChecks',
--- a/index.js
+++ b/index.js
@@ -141,7 +141,7 @@ const featuresToIncludeTempUnprotectedExceptions = [
     'navigatorInterface',
     'nonTracking3pCookies',
     'performanceMetrics',
-    'phishingDetection',
+    'maliciousSiteProtection',
     'referrer',
     'requestFilterer',
     'runtimeChecks',
--- a/overrides/ios-override.json
+++ b/overrides/ios-override.json
@@ -582,17 +582,6 @@
                 }
             }
         },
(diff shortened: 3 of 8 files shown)


not-a-rootkit:
FYI we're still discussing whether we want this much granular control over the feature. TBD Monday.

overrides/ios-override.json
```diff
@@ -582,16 +582,13 @@
                 }
             }
         },
-        "phishingDetection": {
```
jonathanKingston, line 585: This never got rolled out right?

features/malicious-site-protection.json
```diff
@@ -0,0 +1,8 @@
+{
+    "_meta": {
+        "description": "Malicious Site Protection",
```
jonathanKingston, line 3: Should we explain the rename in the phishing protection feature (aka it's deprecated)

index.js
```diff
@@ -142,6 +142,7 @@ const featuresToIncludeTempUnprotectedExceptions = [
     'nonTracking3pCookies',
     'performanceMetrics',
     'phishingDetection',
+    'maliciousSiteProtection',
```
jonathanKingston, line 145: I suspect we never wanted that here. This is why there are "exceptions"

schema/features/malicious-site-protection.ts
```diff
@@ -0,0 +1,11 @@
+import { Feature, CSSInjectFeatureSettings } from '../feature';
+
+type State = 'enabled' | 'disabled' | 'internal';
+type StateObject = {
+    state: State;
+};
```
jonathanKingston, line 6:
```suggestion
```

jonathanKingston: (I think this is unused anyway, looks like we don't have a lint that prevents unused types)

overrides/ios-override.json
not-a-rootkit: Only internally yes.

index.js
not-a-rootkit: I thought we wanted exceptions in general, I just want to be able to define them manually though. If we remove it from here can we still do that in the `overrides/*.json`?

jonathanKingston: nah just means temp allowlisted sites get an exception here. It's for website investigations where we don't know the cause.

We can still use overrides as you say :)

jonathanKingston (approved):

PR #2508, opened 2024-11-22
Revert "Set setAccessTokenCookieForSubscriptionDomains to enabled for iOS and macOS"

samsymons:
Reverts duckduckgo/privacy-configuration#2506

This change is aligned with a large increase in crashes so I'm disabling it while investigating it further, as it's the only remote change that lines up with the spike. I'll re-enable this later today if it turns out to be unrelated.


brindy (approved):
LGTM

PR #2558, opened 2024-12-09
Disable `windowsNewTabPageExperiment`

greyivy:
**Asana Task/Github Issue:** N/A

## Description
Disables the NTP experiment to facilitate a downgrade during the [ongoing incident](https://app.asana.com/0/7643933462314/1208939348486306/f).

#### Reference

-   [Config Reviewer Documentation](https://app.asana.com/0/1200890834746050/1204443212791216/f)
-   [Config Maintainer Documentation](https://app.asana.com/0/1200890834746050/1200573250322769/f)
-   [Feature Implementer Documentation](https://app.asana.com/0/1200890834746050/1201498956177210/f)

PR #2559, opened 2024-12-10
Revert "Disable `windowsNewTabPageExperiment`"

greyivy:
Reverts duckduckgo/privacy-configuration#2558


mikamikem (approved):
Yay and back the way it was not so long ago!

PR #2571, opened 2024-12-11
Further reduce the mitigation for asuracomic.net

kzar:
It seems this is another instance of "surrogate" (aka) shim script redirection
timing issues in the macOS and iOS browsers. Instead of disabling content
blocking entirely, let's just stop redirecting the problematic script.

**Asana Task/Github Issue:** https://app.asana.com/0/1206670747178362/1208879720624568/f


laghee (approved):
WFM!

PR #2575, opened 2024-12-12
Disable 1p-cookie expiry on SERP.

sammacbeth:
**Asana Task/Github Issue:** https://app.asana.com/0/1200123405659223/1207370736410846/f

## Description
Prevents SERP settings from being reset after 7 days.

### DIFF
--- a/features/cookie.json
+++ b/features/cookie.json
@@ -57,6 +57,10 @@
         {
             "domain": "instructure.com",
             "reason": "https://github.com/duckduckgo/privacy-configuration/issues/2212"
+        },
+        {
+            "domain": "duckduckgo.com",
+            "reason": "Prevent expiry of SERP settings"
         }
     ]
 }


kdzwinel (approved):
This makes sense as a temporary measure to reduce user annoyance, but let's keep pushing for a better solution as discussed.

features/cookie.json
```diff
@@ -35,7 +35,12 @@
             "threshold": 604800,
             "maxAge": 604800
         },
-        "thirdPartyCookieNames": ["user_id", "__Secure-3PAPISID", "SAPISID", "APISID"]
+        "thirdPartyCookieNames": [
+            "user_id",
+            "__Secure-3PAPISID",
+            "SAPISID",
+            "APISID"
+        ]
```
kdzwinel, line 43: I think that's what linter is complaining about, you'll need to revert it.

kdzwinel (approved):

PR #2594, opened 2024-12-17
Enable auto-tapping on import password flow when not signed in

CDRussell:
**Asana Task/Github Issue:** https://app.asana.com/0/608920331025315/1208992440597713/f

## Description
Enables the auto-tapping of two buttons inside the import Google Password flow. Right now, we show a UI hint in the form of a button animation in three places:
1. [if user not already signed in] Sign In button on initial landing page
2. [if user wasn’t signed in] After signing in, on the settings cog which is on the screen prior to where they can export
3. On the final export button

This PR converts `#1` and `#2` to auto-tap instead of showing an animation.

### DIFF
--- a/overrides/android-override.json
+++ b/overrides/android-override.json
@@ -429,7 +429,7 @@
                                             "op": "add",
                                             "value": {
                                                 "settingsButton": {
-                                                    "shouldAutotap": false,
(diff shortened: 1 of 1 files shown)


cmonfortep (approved):
LGTM

PR #2688, opened 2025-01-28
Validate that cohort names do not contain special characters

sammacbeth:
**Asana Task/Github Issue:** https://app.asana.com/0/1208889145294658/1208747415972722/f

## Description

Cohort names are passed directly to some pixels, so we should make sure that they don't contain any special characters that could break the analysis pipeline (e.g. `_`, `.` or `?`). This PR adds a unit-test to validate all cohort names in the config against the regex `/^[a-zA-Z0-9]+$/`.

### DIFF
--- a/tests/config-tests.js
+++ b/tests/config-tests.js
@@ -109,6 +109,22 @@ describe('Config schema tests', () => {
                     }
                 }
             });
+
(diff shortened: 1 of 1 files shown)


marcosholgado (approved):
LGTM

jonathanKingston (approved):

PR #2861, opened 2025-03-14
Hide blank space on youtube scroll

laghee:
**Asana Task/Github Issue:** https://app.asana.com/0/1206670747178362/1209662493720640/f

#### Brief explanation
- Reported URL: m.youtube.com
- Problems experienced: blank space where content from ad tracker was blocked
- Platforms affected:
  - [x] iOS
  - [x] Android
  - [ ] Windows
  - [ ] MacOS
  - [ ] Extensions

#### Reference

### DIFF
--- a/features/element-hiding.json
+++ b/features/element-hiding.json
@@ -3589,6 +3589,15 @@
                 ]
             },
             {
+                "domain": "youtube.com",
+                "rules": [
+                    {
+                        "selector": "ytm-promoted-sparkles-web-renderer",
+                        "type": "hide"
+                    }
+                ]
+            },
+            {
                 "domain": "first-party.site",
                 "rules": [
                     {


features/element-hiding.json
```diff
@@ -3588,6 +3588,15 @@
                     }
                 ]
             },
+            {
+                "domain": "youtube.com",
+                "rules": [
+                    {
+                        "selector": "ytm-promoted-sparkles-web-renderer",
+                        "type": "hide"
```
kzar, line 3596: @laghee Did you check if "hide-empty" works? (@dharb mentioned we should prefer hide-empty vs hide where possible)

jonathanKingston: Should we enshrine this in a automated rule? There's a lot of nuance in per repo reviews and I'd like us to automate this knowledge a bit even if just to ask these questions,

kzar: That might be good, I'm not sure exactly how it would work - could it automatically add an inline comment on the PR like I did? If that had happened here, @laghee could have quickly adjusted the PR or could have left a reply to explain why "hide" really was necessary.

_(If that is possible, the automated comment might be something like this: `"hide-empty" is preferred over "hide", since that way elements are only hidden when necessary (e.g. when their contents were blocked). If "hide" is necessary here, please explain why.`._)

laghee: @kzar In this case we're only hiding the ad image, not the ad CTAs, and the element holding the ad image is the same one they use to hold thumbnails for all other videos in the scroll so I didn't see any other way to effectively fix the issue in a privacy-preserving way. The text for both the CTA elements will change depending on which ad doubleclick decides to load there, so no way to filter out by text either. If there's another way to do this effectively, I wasn't aware of it.

italianAdusAd

kzar: Cool, thanks for confirming @laghee 👍

kzar (approved):

PR #2893, opened 2025-03-24
Fix broken interaction on stamma.org

kzar:
It seems after the cookie prompt is automatically handled, scrolling on the
website breaks. So let's disable that for now.

**Asana Task/Github Issue:** https://app.asana.com/0/1206670747178362/1209724647074238/f


dharb (approved):
👍

PR #2900, opened 2025-03-25
Hide empty space in header of US local news websites

kzar:
It seems quite a few US local news websites are similarly structured and have
empty space in the header. Let's make the existing DisplayAdvert rule global so
it applies to all of those websites.

**Asana Task/Github Issue:** https://app.asana.com/0/1206670747178362/1209746380480547/f


dharb (approved):
👍

PR #2904, opened 2025-03-26
Hide more empty space on minkabu.jp

kzar:
There is also some empty space in the sidebar, let's hide that too.

**Asana Task/Github Issue:** https://app.asana.com/0/1204912272578138/1209779991649706/f

### DIFF
--- a/features/element-hiding.json
+++ b/features/element-hiding.json
@@ -2304,6 +2304,10 @@
                     {
                         "selector": ".ad",
                         "type": "closest-empty"
+                    },
+                    {
+                        "selector": "div#side_navi > div:has(> div > div.ad)",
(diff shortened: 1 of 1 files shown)


features/element-hiding.json
```diff
@@ -2304,6 +2304,10 @@
                     {
                         "selector": ".ad",
                         "type": "closest-empty"
+                    },
+                    {
+                        "selector": "div#side_navi > div:has(> div > div.ad)",
+                        "type": "hide"
```
kzar, line 2310: I tried "hide-empty" but that didn't work, even after I tried adding the Japanese "adLabelStrings" array.

tagawa (approved):

PR #2748, opened 2025-02-12
Add exception for d3qpjnspil4z56.cloudfront.net

not-a-rootkit:
**Asana Task/Github Issue:** https://app.asana.com/0/1208859185135959/1209357181668521

## Description
We are getting reports that apps.facebook.com/texas_holdem is causing breakage due to it loading an iframe that is flagged as malicious in our dataset. 

### Feature change process:

- [ ] I have added a [schema](https://github.com/duckduckgo/privacy-configuration/tree/main/schema) to validate this feature change
- [x] I have tested this change locally
- [x] This code for the config change is ready to merge
- [x] This feature was covered by a tech design

#### Additional info:

- [x] I have tested this change locally in all supported browsers
- [x] This change will be visible to users
- [x] This code for the config change is ready
- [ ] This change was covered by a ship review

#### Brief explanation
- Reported URL: https://apps.facebook.com/texas_holdem
- Problems experienced: Malicious site protection warning page is shown and since it's embedded in an iframe that redirects, it causes users to get stuck in a loop unable to bypass the error page.
- Platforms affected:
  - [ ] iOS
  - [ ] Android
  - [ ] Windows
  - [x] MacOS
  - [ ] Extensions
- Tracker(s) being unblocked: None.
- Feature being disabled: None, just adding exception for domain

- [ ] I have referenced the URL of this PR as the "reason" value for the exception (where applicable).
- [ ] This change is a speculative mitigation to fix reported breakage.

#### Reference

-   [Config Reviewer Documentation](https://app.asana.com/0/1200890834746050/1204443212791216/f)
-   [Config Maintainer Documentation](https://app.asana.com/0/1200890834746050/1200573250322769/f)
-   [Feature Implementer Documentation](https://app.asana.com/0/1200890834746050/1201498956177210/f)

PR #2912, opened 2025-03-27
Add script to remove stale exceptions for malicious site protection

not-a-rootkit:
**Asana Task/Github Issue:**

## Description
As part of [Malicious Site Protection Breakage: Triage Playbook](https://app.asana.com/0/0/1209432061170256) we are required to add new entries into the [privacy-config](https://github.com/duckduckgo/privacy-configuration/pull/2748) to exclude domains from our checks. After a certain amount of time, this may grow uncontrollably, and impact (A) the size of our config, and (B) the performance of our checks. 

Therefore, we have this process which is a core responsibility of  [Malicious Site Protection Triage Rotation](https://app.asana.com/0/0/1209785731259891), where we must remove entries that are no longer needed from the config. Thankfully, this is a fairly easy process, and can be partially automated with a script in the repo.

### To test
- Clone privacy-configuration repository
- Add an exclusion for a random domain to one of either the overrides or default configs, e.g. in features/malicious-site-protection.json
  - Add { "domain": "fakesite.com", "reason": "example" }
- Run npm run build  to build the configs
- Run node scripts/filter-malicious.js
- This script will check all exclusions against the API and remove exclusions that are no longer needed, in this case it should remove this domain from the default config
- Now test the overrides, e.g. in overrides/macos-override.json, under features.maliciousSiteProtection.exclusions  add it again
- Run npm run build
- Run node scripts/filter-malicious.js
- Ensure it has removed it from the right override.

There is one side-effect of this script, it will force standardised formatting to our JSON that is slightly different to what we use now. For example:

```
// This...
-                                "domains": ["<all>"],
// Will become...
+                                "domains": [
+                                    "<all>"
+                                ],
```

I could avoid this, but I don't think it can be done without a custom JSON encoder which is overkill for a quick script that we run once a month.

### Feature change process:

*** N/A this is just a script that right now doesn't affect any existing configs ***

#### Additional info:

#### Brief explanation
- Reported URL:
- Problems experienced:
- Platforms affected:
  - [ ] iOS
  - [ ] Android
  - [ ] Windows
  - [ ] MacOS
  - [ ] Extensions
- Tracker(s) being unblocked:
- Feature being disabled:

- [ ] I have referenced the URL of this PR as the "reason" value for the exception (where applicable).

#### Reference


jonathanKingston:
> I could avoid this, but I don't think it can be done without a custom JSON encoder which is overkill for a quick script that we run once a month.

You should run prettier on it afterwards. We're likely to change the formatting soon of this inline array thing but running prettier after should make it pass the tests we already have.

scripts/filter-malicious.js
```diff
@@ -0,0 +1,151 @@
...
+    */
+    async updateDefaultConfig(updatedExceptions) {
+        const defaultConfig = await this.loadConfig(this.defaultConfig);
+        if (defaultConfig?.exceptions) {
+            const updatedDefaultExceptions = defaultConfig.exceptions.filter((exception) => {
+                return updatedExceptions.some((updated) => updated.domain === exception.domain);
+            });
+            if (updatedDefaultExceptions.length !== defaultConfig.exceptions.length) {
+                defaultConfig.exceptions = updatedDefaultExceptions;
+                await this.saveConfig(this.defaultConfig, defaultConfig);
```
jonathanKingston, line 137: Should we consider just running this every week and assigning to a group of folk to review if there are changes?

Optionally/Additionally should we put a added date to the exceptions? That way we could ignore exceptions that were added within a certain time period? (and potentially fully automate this).

not-a-rootkit: Not sure if adding a date filter adds that much more value, but I do like the idea of automating the whole thing. I had a (painful) crack at that here, you can see the autogenerated PR here: https://github.com/duckduckgo/privacy-configuration/pull/2920.

Once I get multiline PR bodies working properly we can give it another look.

not-a-rootkit: This is better:  https://github.com/duckduckgo/privacy-configuration/pull/2921

jonathanKingston (approved):

PR #2934, opened 2025-04-02
Prevent parallel asana sync actions

sammacbeth:
**Asana Task/Github Issue:**

## Description
Multiple review requested workflows triggered in parallel can cause duplicate tasks to be created. Let's add concurrency control to the workflow to prevent that.


jonathanKingston (approved):

jonathanKingston:
Nice TIL!

PR #2940, opened 2025-04-03
Prevent feature automatic overriding going forward

jonathanKingston:
**Asana Task/Github Issue:**

### Feature change process:

#### Additional info:

#### Brief explanation
- Reported URL:
- Problems experienced:
- Platforms affected:
  - [ ] iOS
  - [ ] Android
  - [ ] Windows
  - [ ] MacOS
  - [ ] Extensions
- Tracker(s) being unblocked:
- Feature being disabled:

- [ ] I have referenced the URL of this PR as the "reason" value for the exception (where applicable).

#### Reference

### DIFF
--- a/index.js
+++ b/index.js
@@ -168,6 +168,61 @@ function isFeatureMissingState(feature) {
     return !('state' in feature);
 }
 
+// We previously rolled out all features to all platforms, so this is a safety whilst we deprecate that behavior.
+// Don't add new items to this list, we should just add fearures to overrides/ files instead with explicit state.
+const legacyDisabledFeatures = [
+    'androidBrowserConfig',
+    'androidNewStateKillSwitch',
+    'ampLinks',
+    'appTrackerProtection',
+    'auraExperiment',
+    'autoconsent',
+    'bookmarksSorting',
+    'brokenSiteReportExperiment',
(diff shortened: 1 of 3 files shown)


overrides/extension-override.json
```diff
@@ -147,6 +147,9 @@
                     }
                 }
             }
+        },
+        "webCompat": {
+            "state": "disabled"
```
jonathanKingston, line 152: This isn't really needed but also it's the only place we're not explicitly enabling webCompat.

jonathanKingston:
I think we should just remove all features from this behavior with v5. I'll keep this open just in case we decide to adopt soon.

graeme (approved):
Looks reasonable to me

index.js
```diff
@@ -242,9 +304,6 @@ async function buildPlatforms() {
             }
 
             platformConfig.features[key] = { ...platformOverride.features[key] };
-            if (isFeatureMissingState(platformConfig.features[key])) {
```
jonathanKingston, line 245: iirc this gets caught with schema now.
If the feature doesn't have a "state" in the features file and also not in the platform. Also it'd be deleted above.

bwaresiak (approved):

PR #3035, opened 2025-04-23
Check that overriden features aren't missing subfeatures

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/task/1210042906543794?focus=true

### Feature change process:

#### Additional info:

#### Reference

### DIFF
--- a/tests/config-tests.js
+++ b/tests/config-tests.js
@@ -152,26 +152,28 @@ describe('Config schema tests', () => {
 });
 
 describe('Config schema tests', () => {
-    const baseFeatures = getBaseFeatureConfigs();
-    for (const platform of platforms) {
-        const overrideConfig = JSON.parse(fs.readFileSync(path.join(__dirname, `/../overrides/${platform}-override.json`), 'utf-8'));
-        // Skip over extension platforms:
-        if (platform.startsWith('browsers/')) {
-            continue;
-        }
-        for (const [featureName, baseFeature] of Object.entries(baseFeatures)) {
-            const overrideFeature = overrideConfig.features[featureName];
-            // Skip over if we have no override for this feature
-            if (!overrideFeature) continue;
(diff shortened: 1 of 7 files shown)


util.js
```diff
@@ -56,6 +58,24 @@ function addPathRule(rules, rule) {
     }
 }
 
+function getBaseFeatureConfigs() {
+    const features = {};
+    // Grab all exception lists
+    const jsonListNames = fs.readdirSync(LISTS_DIR).filter((listName) => {
+        return listName !== UNPROTECTED_LIST_NAME && listName !== '_template.json';
+    });
```
sammacbeth, line 66: I'd also filter for `.json` suffix, just in case other hidden or cache files appear in that dir (e.g. when working with this locally).

Edit: I now see that this is just moving existing code, so that refactor suggestion is optional :)

jonathanKingston: > Edit: I now see that this is just moving existing code, so that refactor suggestion is optional :)

Let me raise an issue for it, I'd rather not block here if you're ok with that. I can resolve tomorrow.

jonathanKingston: Raised: https://app.asana.com/1/137249556945/project/1200890834746050/task/1210163485530445?focus=true

tests/config-tests.js
```diff
@@ -147,3 +150,28 @@ describe('Config schema tests', () => {
         });
     }
 });
+
+describe('Config schema tests', () => {
```
sammacbeth, line 154: You've got the test in the `describe` block, instead of the `it` block. Does it still work?

I think we can just add an `it` with a description of what we're testing for, e.g.
```javascript
describe('Platform overrides', () => {
  it('each overriden feature should include all sub features from base feature', () => {
```

sammacbeth (changes requested):

sammacbeth (approved):

PR #3076, opened 2025-05-02
Fix broken page interaction on riga-airport.com

kzar:
The Cookie Prompt Management (CPM) feature seems to clash with riga-airport.com,
leaving the website in a broken state. When this happens, there's a grey overlay
and page interaction (e.g. clicking) doesn't work.

**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1206670747178362/task/1210107510466911?focus=true


laghee (approved):
👍

PR #3114, opened 2025-05-08
Enable Sense Of Protection Experiment

malmstein:
**Asana Task/Github Issue:**https://app.asana.com/1/137249556945/project/1174433894299346/task/1210140711789469?focus=true

## Description
<!-- 
  Please delete either or both process sections below.
—>
Enables the Sense Of Protection experiment for O-10

### Feature change process:

#### Brief explanation
- Reported URL:
- Problems experienced:
- Platforms affected:
  - [ ] iOS
  - [ ] Android
  - [ ] Windows
  - [ ] MacOS
  - [ ] Extensions
- Tracker(s) being unblocked:

### DIFF
--- a/features/tile-switcher-animation.json
+++ b/features/tile-switcher-animation.json
@@ -1,0 +1,7 @@
+{
+    "_meta": {
+        "description": "A feature flag that determines whether the tile switcher info panel animation is enabled.",
+        "sampleExcludeRecords": {}
+    },
+    "exceptions": []
+}
--- a/features/sense-of-protection.json
+++ b/features/sense-of-protection.json
@@ -1,0 +1,7 @@
+{
+    "_meta": {
+        "description": "Controls the Sense Of Protection experiment."
+    },
+    "state": "disabled",
+    "exceptions": []
+}
--- a/overrides/android-override.json
+++ b/overrides/android-override.json
@@ -2092,6 +2092,54 @@
                     ]
                 }
             }
+        },
+        "tileSwitcherAnimation": {
+            "state": "enabled",
+            "minSupportedVersion": 52320000
+        },
+        "senseOfProtection": {
+            "state": "enabled",
+            "exceptions": [],
+            "minSupportedVersion": 52320000,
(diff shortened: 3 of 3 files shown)


features/tile-switcher-animation.json
```diff
@@ -0,0 +1,7 @@
+{
+    "_meta": {
+        "description": "A feature flag that determines whether the tile switcher info panel animation is enabled.",
```
jonathanKingston, line 3: This seems like a sub feature, does it need to be a parent one?

marcosholgado (dismissed):
LGTM

features/tile-switcher-animation.json
marcosholgado: Actually it's a parent. The name of the json though should probably be tab-switcher-animation.json rather than `tile`

features/tab-switcher-animation.json
```diff
@@ -0,0 +1,7 @@
+{
+    "_meta": {
+        "description": "A feature flag that determines whether the tab switcher info panel animation is enabled.",
+        "sampleExcludeRecords": {}
```
marcosholgado, line 4: I don't think this needs this samplee xcluded records? Also missing the default state disabled

malmstein: sorted

marcosholgado (approved):
LGTM

features/tile-switcher-animation.json
malmstein: sorted

malmstein:
@jonathanKingston can you review this again?

features/tile-switcher-animation.json
jonathanKingston: My question is why isn’t it a sub feature?

jonathanKingston: I'll add to my guidance doc, ideally we should be raising draft config ahead of time and testing it in the client too.

jonathanKingston (approved):

PR #3141, opened 2025-05-12
Enable remote broker delivery for internal testing

quanganhdo:
**Asana Task/Github Issue:**

https://app.asana.com/1/137249556945/project/72649045549333/task/1205508328452434?focus=true

## Description
<!-- 
  Please delete either or both process sections below.
—>

Enables PIR remote broker delivery for internal users.

### Feature change process:

#### Brief explanation
- Reported URL:
- Problems experienced:
- Platforms affected:
  - [ ] iOS
  - [ ] Android
  - [ ] Windows
  - [ ] MacOS
  - [ ] Extensions
- Tracker(s) being unblocked:

### DIFF
--- a/overrides/windows-override.json
+++ b/overrides/windows-override.json
@@ -76,9 +76,6 @@
                 },
                 "freemium": {
                     "state": "disabled"
-                },
-                "remoteBrokerDelivery": {
-                    "state": "disabled"
                 }
             },
             "settings": {
--- a/overrides/windows-override.json
+++ b/overrides/windows-override.json
@@ -76,6 +76,9 @@
                 },
                 "freemium": {
                     "state": "disabled"
+                },
+                "remoteBrokerDelivery": {
+                    "state": "disabled"
                 }
             },
             "settings": {
--- a/overrides/windows-override.json
+++ b/overrides/windows-override.json
@@ -158,6 +158,7 @@
         },
         "duckPlayer": {
             "state": "enabled",
+            "minSupportedVersion": "0.109.7",
             "features": {
                 "pip": {
                     "state": "enabled"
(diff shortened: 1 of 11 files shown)


samsymons (dismissed):
The macOS override looks good to me!

overrides/windows-override.json
```diff
@@ -76,6 +76,9 @@
                 },
                 "freemium": {
                     "state": "disabled"
+                },
+                "remoteBrokerDelivery": {
+                    "state": "disabled"
```
quanganhdo, line 81: Adding this satisfies the "All subfeatures must be defined in their overrides files if they apply” CI check.

Not sure it’s the right way to do it, as technically Windows supports remote broker delivery but it’s not behind any feature flag.

cc @jonathanKingston

jonathanKingston: I think making it "disabled" makes the least confusion for it's current state.

We could just make it macos specific? So not defined in the feature file?

q71114: We already have these two features:
- https://github.com/duckduckgo/privacy-configuration/blob/main/features/broker-protection.json: used in C-S-S
- https://github.com/duckduckgo/privacy-configuration/blob/main/features/dbp.json: used in Windows/macOS for waitlist
Can we add this as a sub-feature to the latter?

quanganhdo: I already did, but CI wants it to be in windows-override.json too https://github.com/duckduckgo/privacy-configuration/blob/634b7e2c5de6d3768ded853329b253b72cea9f43/features/dbp.json#L16-L18

q71114: In this case I agree with @jonathanKingston, defining it only in `macos-override.json` would be the least confusing if it's possible.

quanganhdo: Sounds good, I’ve removed these other changes. @jonathanKingston Back to you for final approval (I think).

q71114: Now it's only a change in `macos-override.json`, this PR can be approved by any Apple dev.

afterxleep (approved):

PR #3195, opened 2025-05-20
Skip disabled exceptions

jonathanKingston:
**Asana Task/Github Issue:**

## Description

Skip delivering exceptions in the config when they're not needed as the feature is disabled.

### Feature change process:

#### Brief explanation
- Reported URL:
- Problems experienced:
- Platforms affected:
  - [ ] iOS
  - [ ] Android
  - [ ] Windows
  - [ ] MacOS
  - [ ] Extensions
- Tracker(s) being unblocked:

### DIFF
--- a/index.js
+++ b/index.js
@@ -259,6 +259,13 @@ async function buildPlatforms() {
             }
         }
 
+        // Skip the config weight of exceptions for features that are disabled
+        for (const [key, feature] of Object.entries(platformConfig.features)) {
+            if (feature.state === 'disabled') {
+                feature.exceptions = [];
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


dharb (changes requested):
I like this idea in theory, but I think it's going to run into issues with features that are disabled by default, but enabled via an experiment - eg this is going to drop all `fingerprintingCanvas` exceptions.

jonathanKingston:
Ah thanks for clarification, for currently disabled features this breaks adding exceptions to them.

I'm going to close and open a discussion as I think we should change how we do that; at least to find a way to make this work. Ideally all disabled features should be safe to remove.

PR #3204, opened 2025-05-21
Remote CPM rules on Android

sammacbeth:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1163321984198618/task/1210318211455596

## Description

- Move rules into the autoconsent feature file.
- Exclude rules on iOS, extension and Windows via platform overrides.


muodov (approved):
Ok so in the current state, compact rules will effectively ship on macos and android, and not on windows/ios.

muodov (approved):

PR #3210, opened 2025-05-22
Improve airlive.net element-hiding exception

sammacbeth:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1206670747178362/task/1210277365867272?focus=true

## Description
Instead of full disabling element-hiding, we can disable generic rules and only target relevant elements on the page.


dharb (approved):
👍

PR #3229, opened 2025-05-27
Add prettier multi-line plugin

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1200890834746050/task/1209117386771959?focus=true

### Feature change process:

#### Brief explanation
- Reported URL:
- Problems experienced:
- Platforms affected:
  - [ ] iOS
  - [ ] Android
  - [ ] Windows
  - [ ] MacOS
  - [ ] Extensions
- Tracker(s) being unblocked:

### DIFF
--- a/.github/scripts/diff-directories.js
+++ b/.github/scripts/diff-directories.js
@@ -13,7 +13,10 @@ function readFilesRecursively(directory) {
 
         if (fileStats.isDirectory()) {
             const nestedFiles = readFilesRecursively(filePath);
-            for (const [nestedFilePath, nestedFileContent] of Object.entries(nestedFiles)) {
+            for (const [
+                nestedFilePath,
+                nestedFileContent,
+            ] of Object.entries(nestedFiles)) {
                 files[path.join(filename, nestedFilePath)] = nestedFileContent;
             }
         } else {
@@ -64,7 +67,10 @@ function displayDiffs(dir1Files, dir2Files, isOpen) {
         rollupGrouping[summary].files.push(fileName);
         rollupGrouping[summary].string = string;
     }
-    for (const [filePath, fileContent] of Object.entries(dir1Files)) {
+    for (const [
+        filePath,
+        fileContent,
+    ] of Object.entries(dir1Files)) {
         let diffOut = '';
         let compareOut;
         if (filePath in dir2Files) {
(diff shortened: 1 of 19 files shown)


jonathanKingston:
@sammacbeth this is weird to me that `ts-json-schema-generator` is broken by the inclusion of `prettier-plugin-multiline-arrays`.

- It's weird to me that a non included module, leaks into the test and causes errors such as [1]. It's even more confusing in that in both modules I don't see the use of `Intl` or `DateTimeFormatPart` member.
- This is what stalled https://github.com/duckduckgo/privacy-configuration/pull/2948.


[1] example error:
```
      33) should validate against the legacy schema
    v3/extension-edg-config.json
      ✔ should contain appTrackerProtection or not
{
  category: 1,
  file: undefined,
  length: 0,
  start: 0,
  code: 'J - 108',
  messageText: 'Type check error',
  relatedInformation: [
    {
      file: [SourceFileObject],
      start: 58997,
      length: 18,
      messageText: "'Intl' has no exported member named 'DateTimeFormatPart'. Did you mean 'DateTimeFormat'?",
      category: 1,
      code: 2724,
      reportsUnnecessary: undefined,
      reportsDeprecated: undefined
    }
  ]
}
```

Note: I've not done the formatting to make the diff much easier to reason with here. This formatting change is causing diff noise pretty regularly, so I'd like to get this plugin enabled. /cc @dharb

tests/schema-validation.js
```diff
@@ -6,6 +6,7 @@ function createGenerator() {
     try {
         return schemaGenerator.createGenerator({
             path: './schema/config.ts',
+            tsconfig: './tsconfig.json',
```
jonathanKingston, line 9: This fixed the mentioned issue, but I think we should discuss if it's safe. And/or if we should create a separate tsconfig.

jonathanKingston:
> Note: ~I've not done the formatting to make the diff much easier to reason with here.~ (as I fixed the test I added it) This formatting change is causing diff noise pretty regularly, so I'd like to get this plugin enabled. /cc @dharb

As the inclusion of the types file fixed it, I added the formatting change. We should still discuss if it's safe enough, I'm not a huge fan of side effects like this.

PR #3241, opened 2025-05-28
Enabled message bridge feature on windows from 0.115.0 onwards

Jonathan-L:
https://app.asana.com/1/137249556945/project/72649045549333/task/1209556560924114?

## Description
This PR enables the `Message Bridge` feature for windows. The first (and currently only) feature that uses this will be released in v0.115.0.

### Feature change process:

### DIFF
--- a/overrides/windows-override.json
+++ b/overrides/windows-override.json
@@ -323,7 +323,7 @@
         "messageBridge": {
             "exceptions": [],
             "settings": {
-                "aiChat": "enabled",
+                "aiChat": "disabled",
                 "domains": [
                     {
                         "domain": "duckduckgo.com",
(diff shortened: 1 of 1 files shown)


overrides/windows-override.json
```diff
@@ -320,6 +320,26 @@
                 }
             }
         },
+        "messageBridge": {
+            "exceptions": [],
+            "settings": {
+                "aiChat": "enabled",
```
jonathanKingston, line 326: I think this should be "disabled" here, the patchSettings then makes this enabled for duckduckgo.com

Jonathan-L: Thanks for the spot! I've changed this now.

PR #3262, opened 2025-06-03
Disable swiping tabs for external users

CrisBarreiro:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/414730916066338/task/1210448666162072?focus=true

## Description
Temporarily disable swiping tabs for external users to validate if that's causing an increase in page load times


cmonfortep (approved):
lgtm

PR #3328, opened 2025-06-18
Android: Re-enable tab swiping

0nko:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1207418217763355/task/1210580640372907?focus=true

## Description

This PR re-enables the tab swiping after the 2 issues related to the feature were fixed, which will be rolled out with version 5.239.0 (min. version for this feature flag).

### DIFF
--- a/overrides/android-override.json
+++ b/overrides/android-override.json
@@ -28277,11 +28277,11 @@
             }
         },
         "swipingTabs": {
-            "state": "disabled",
-            "minSupportedVersion": 52380000,
+            "state": "enabled",
+            "minSupportedVersion": 52390000,
             "features": {
                 "enabledForUsers": {
-                    "state": "disabled"
+                    "state": "enabled"
                 }
             }
         },


overrides/android-override.json
```diff
@@ -28277,11 +28277,11 @@
...
         },
         "swipingTabs": {
-            "state": "disabled",
-            "minSupportedVersion": 52380000,
+            "state": "enabled",
+            "minSupportedVersion": 52390000,
             "features": {
                 "enabledForUsers": {
-                    "state": "disabled"
+                    "state": "enabled"
```
LukasPaczos, line 28284:
```suggestion
                    "state": "enabled",
                    "minSupportedVersion": 52390000
```

I know that in the [integration code for this specific feature](https://github.com/duckduckgo/Android/blob/6a27ab29e43f369db36737b1ffd149584cba6653/common/common-ui/src/main/java/com/duckduckgo/common/ui/tabs/SwipingTabsFeatureProvider.kt#L28) parent and sub-feature flags are combined using an `AND` clause, but since each flag is interpreted independently by the Android feature flag system, it's best to specify all flag parameters separately, so that the state is clear just by reading the config file.

@jonathanKingston @aitorvs @marcosholgado this case might be an interesting example for future conversations related to https://app.asana.com/1/137249556945/project/1208671518894266/task/1210445354192644?focus=true and config rules/best practices.

0nko: I can't commit the suggestion for some reason. I added this in a separate commit. @LukasPaczos, can you please review?

LukasPaczos (approved):

PR #3378, opened 2025-06-30
Increase the Visual Refresh on Windows to 25%

anpete:
https://app.asana.com/1/137249556945/project/72649045549333/task/1210176770043635?focus=true

## Description
Increasing the rollout of the Visual Refresh on Windows to 25% in the stable channel.

### DIFF
--- a/overrides/windows-override.json
+++ b/overrides/windows-override.json
@@ -768,7 +768,7 @@
                     "rollout": {
                         "steps": [
                             {
-                                "percent": 5
+                                "percent": 25
                             }
                         ]
                     }


overrides/windows-override.json
```diff
@@ -768,7 +768,7 @@
                     "rollout": {
                         "steps": [
                             {
-                                "percent": 5
+                                "percent": 25
```
q71114, line 774: For the rollout code to work properly you need to add a separate step for each increase. (The code checks for number of steps.)

ranrande: Documented here: https://app.asana.com/1/137249556945/project/1208736637614995/task/1207579150090376?focus=true

anpete: Ok, thanks!

q71114 (approved):

PR #3385, opened 2025-06-30
Add GHA to check invalid config for contentScopeExperiments (Android)

CrisBarreiro:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/task/1210674581657609?focus=true

## Description
Add a GitHub action that will
* Fail and create a task if a new experiment is added and enabled in contentScopeExperiments and enabled for Android and ContentScopeExperimentsFeature on the Android codebase doesn't have it defined
* Pass and create a task if less than 10 of the pre-allocated experiments for contentScopeExperiments in the Android codebase are left

### DIFF
--- a/.github/workflows/check-android-content-scope-experiments.yml
+++ b/.github/workflows/check-android-content-scope-experiments.yml
@@ -1,0 +1,126 @@
+name: Check Latest Content Scope Experiment State
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'features/content-scope-experiments.json'
+      - 'overrides/android-override.json'
+  pull_request:
+    paths:
+      - 'features/content-scope-experiments.json'
+      - 'overrides/android-override.json'
+
+jobs:
+  check-experiment-state:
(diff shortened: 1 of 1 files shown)


.github/workflows/check-android-content-scope-experiments.yml
```diff
@@ -0,0 +1,126 @@
...
+
+                  {
+                    echo 'task_description<<EOF'
+                    echo -e "$DESCRIPTION"
+                    echo 'EOF'
+                  } >> "$GITHUB_OUTPUT"
+
+            - name: Create Asana Task
+              if: steps.asana_content.outcome == 'success'
+              uses: honeycombio/gha-create-asana-task@main
```
jonathanKingston: Can we use https://github.com/duckduckgo/native-github-asana-sync which is already used on the repo and made internally?

CrisBarreiro: Done

PR #3428, opened 2025-07-08
Revert GPC exclusion for npr.org

laghee:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1206670747178362/task/1207848824036080?focus=true

## Description
NPR was not showing a donation modal when detecting the GPC flag, but not adjusting the layout to account for that, which disabled clicks for users. Seems to be fixed on their end now, so rolling back the exclusion.


dharb (approved):
👍

PR #3514, opened 2025-07-30
SBT add cpm and element hiding mitigation to remove adblock wall

icodebyamanda:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1206670747178362/task/1210924466205327?focus=true

#### Brief explanation
- Reported URL: https://www.kleinezeitung.at/international/19949244/starkes-erdbeben-vor-russischer-halbinsel-kamtschatka-tsunami-welle-usa
- Problems experienced: ad block wall banner prevents access to the site. Mitigation requires to turn off cpm and hide-empty a selector
- Platforms affected:
  - [x ] iOS
  - [ x] Android
  - [x ] Windows
  - [x ] MacOS
  - [ ] Extensions
- Tracker(s) being unblocked: NA
- Feature being disabled/modified: autoconsent and elementHiding

### DIFF
--- a/features/element-hiding.json
+++ b/features/element-hiding.json
@@ -2205,7 +2205,7 @@
                         "type": "closest-empty"
                     }
                 ]
-            }, 
+            },
             {
                 "domain": "kleinezeitung.at",
                 "rules": [


laghee:
Why is this auto-approving something that's failing tests? Is that just something we need to add or are those meant to be separate?

laghee (dismissed):
Not sure why lint is mad — could be whitespace distinction, maybe?

laghee (approved):

jonathanKingston:
“Why is this auto-approving something that's failing tests? Is that just something we need to add or are those meant to be separate?”

IMO fine to be separate, even with an approved review it’ll not merge.

The approval is based on if a human should be helping check it further.

Just run `npm run lint-fix` to fix whatever the problem is. We’ve discussed auto running this too as it’s a bit annoying.

PR #3565, opened 2025-08-07
Add feature flag for new feedback form

jotaemepereira:
**Asana Task/Github Issue:**
https://app.asana.com/1/137249556945/project/1204006570077678/task/1210733970843924?focus=true

## Description
Adds the feature flag to show the new feedback forms. Disabled on all platforms, but enabled on macOS.

### Feature change process:

#### Brief explanation
- Reported URL:
- Problems experienced:
- Platforms affected:
  - [ ] iOS
  - [ ] Android
  - [ ] Windows
  - [x] MacOS
  - [ ] Extensions
- Tracker(s) being unblocked:


overrides/macos-override.json
```diff
@@ -718,6 +718,9 @@
         "disableFireAnimation": {
             "state": "enabled"
         },
+        "newFeedbackForm": {
```
jonathanKingston: I don't think we need a features/ file here (as it's a short lived flag that's not going to be shared).

Can you add `"exceptions": []` here and then we should pass the tests to get this shipped as macOS only.

bwaresiak: I agree - if this is macOS only, let's keep it to macOS, unless you know there's going to be Windows counterpart soon (tho - it can be also implemented in config when needed)

jotaemepereira: @bwaresiak @jonathanKingston This will not be only macOS. Windows will use it too. Their project is starting soon.

jonathanKingston: Discussed here, we're dropping the "new" https://app.asana.com/1/137249556945/task/1210733970843924?focus=true and setting a task reminder to drop if no longer used as well.

jonathanKingston (approved):

samsymons (approved):

PR #3636, opened 2025-08-25
Add web view feature and web view provider windows sub feature

szanto90balazs:
**Asana Task/Github Issue:**

https://app.asana.com/1/137249556945/project/72649045549333/task/1211008971469490?focus=true

## Description

Adds the web view feature. Also adds the web view provider sub feature to the windows-override.

### Feature change process:

### DIFF
--- a/overrides/windows-override.json
+++ b/overrides/windows-override.json
@@ -828,6 +828,14 @@
                 }
             }
         },
+        "webviewProvider": {
+            "state": "enabled",
+            "features": {
+                "webViewProviderPreviewOptIn": {
+                    "state": "disabled"
+                }
+            }
+        },
         "remoteMessaging": {
             "state": "enabled"
         },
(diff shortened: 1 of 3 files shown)


overrides/windows-override.json
```diff
@@ -828,6 +828,14 @@
                 }
             }
         },
+        "webviewProvider": {
```
q71114: Shouldn't we just add a `webview` feature?

szanto90balazs: I agree, made the change 👍 It also makes the top / sub level feature naming more clear

overrides/windows-override.json
```diff
@@ -828,6 +828,14 @@
                 }
             }
         },
+        "webview": {
```
jonathanKingston, line 831: Add exceptions here and you can delete the features/ file and not need a config owners review. It's only for cross platform.

szanto90balazs: Thanks for the recommendation, I didn't consider I could just do this way, makes more sense.

jonathanKingston (approved):

PR #3639, opened 2025-08-25
Enable 25% August TDS

GuiltyDolphin:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1200342317600122/task/1211137585602002?focus=true

### DIFF
--- a/overrides/macos-override.json
+++ b/overrides/macos-override.json
@@ -274,18 +274,18 @@
                 }
             ],
             "features": {
-                "tdsNextExperimentAug25": {
-                    "state": "disabled",
+                "tdsNextExperiment002": {
+                    "state": "enabled",
                     "rollout": {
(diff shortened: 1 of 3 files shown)


overrides/macos-override.json
```diff
@@ -274,18 +274,18 @@
...
             ],
             "features": {
-                "tdsNextExperimentAug25": {
-                    "state": "disabled",
+                "tdsNextExperiment002": {
+                    "state": "enabled",
                     "rollout": {
                         "steps": [
                             {
-                                "percent": 100
```
SabrinaTardio, line 282: Just a note that when we want to increase the percentage we need to add a step (not just increasing the number)
Just double checking cause here I see a single 100% step which doesn’t make sense….

PR #3666, opened 2025-08-28
End pop-under experiment

dharb:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/72649045549333/task/1210346598881910?focus=true

## Description
Ends pop-under experiment. We are going to assess results and decide if we want to go in this direction.


overrides/ios-override.json
```diff
@@ -407,36 +407,6 @@
                             "weight": 1
                         }
                     ]
-                },
-                "scriptletExperiment": {
-                    "state": "enabled",
```
jonathanKingston, line 412: I think this is supposed to remain. Just set it as disabled for now.

PR #3670, opened 2025-08-28
Change GITHUB_TOKEN to PRIVACY_CONFIG_PAT

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1200890834746050/task/1210884021816459?focus=true

## Description

Switch to out friendly neighbourhood duck to do the approving so it can be in the codeowners.

### Feature change process:

#### Brief explanation
- Reported URL:
- Problems experienced:
- Platforms affected:
  - [ ] iOS
  - [ ] Android
  - [ ] Windows
  - [ ] MacOS
  - [ ] Extensions
- Tracker(s) being unblocked:

### DIFF
--- a/.github/workflows/auto-respond-pr.yml
+++ b/.github/workflows/auto-respond-pr.yml
@@ -178,7 +178,7 @@ jobs:
                   gh pr review --approve "$PR_URL"
               env:
                   PR_URL: ${{ github.event.pull_request.html_url }}
-                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


bwaresiak:
@jonathanKingston while I don't mind overrides, I wonder about features - we do catch some misuse from time to time (e.g. using features for single-platform flags). It's not a strong concern tho - wdyt?

jonathanKingston:
@bwaresiak it's not a straight allow, we only allow certain features in any of the changes: https://github.com/duckduckgo/privacy-configuration/blob/main/automation-utils.js#L10-L36

Overrides has the same amount of risk imo, that's why it's laser focused only on exception type changes for now.

PR #3731, opened 2025-09-10
Enable pinned tabs in preview from 0.129

ranrande:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1210132634900450/task/1211305266912265?focus=true

### Feature change process:

### DIFF
--- a/overrides/windows-override.json
+++ b/overrides/windows-override.json
@@ -1177,7 +1177,8 @@
             }
         },
         "pinnedTabs": {
-            "state": "internal",
+            "state": "preview",
+            "minSupportedVersion": "0.129.0",
             "features": {}
         },
         "appHealth": {


overrides/windows-override.json
```diff
@@ -1177,7 +1177,8 @@
             }
         },
         "pinnedTabs": {
-            "state": "internal",
+            "state": "preview",
```
anpete, line 1180: Are you planning an incremental rollout of pinned tabs? If so, you would need to use a sub-feature...

ranrande: I'm not sure yet, the plan is to leave them in preview and monitor their usage and/or user reports, then decide when/how to release to stable. If we need incremental rollout I'll do it at that point.

PR #3929, opened 2025-10-15
rework and add schema validation for element hiding feature

icodebyamanda:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/72649045549333/task/1211479926220518?focus=true

## Description

add element hiding schema definition and register element hiding feature in main config schema

### Feature change process:

---

> [!NOTE]
> Add `ElementHiding` feature schema and include it in `ConfigV5.features`.
> 
> - **Schema**:
>   - **New**: Define `ElementHidingFeature` in `schema/features/element-hiding.ts` with `rules`, domain-specific `rules`, timeouts, and auxiliary selectors/labels.
>   - **Config**: Import and expose `elementHiding?: ElementHidingFeature<VersionType>` in `schema/config.ts` under `ConfigV5.features`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit da3642a56fd0e8103c4fb289d441e7f6e09c34f0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/schema/features/element-hiding.ts
+++ b/schema/features/element-hiding.ts
@@ -22,6 +22,7 @@ export type ElementHidingFeature<VersionType> = Feature<
                 }>;
             }>;
         }>;
+        hideTimeouts?: number[];
         unhideTimeouts?: number[];
         mediaAndFormSelectors?: string;
         adLabelStrings?: string[];


schema/features/element-hiding.ts
```diff
@@ -0,0 +1,31 @@
...
+                values?: Array<{
+                    property: string;
+                    value: string;
+                }>;
+            }>;
+        }>;
+        unhideTimeouts?: number[];
+        mediaAndFormSelectors?: string;
+        adLabelStrings?: string[];
+        [key: string]: any;
```
laghee, line 29: Is this a typo? Maybe meant to be `hideTimeouts`?

```suggestion
        hideTimeouts?: number[];
```

schema/features/element-hiding.ts
```diff
@@ -0,0 +1,31 @@
+import { Feature } from '../feature';
+
+export type ElementHidingFeature<VersionType> = Feature<
+    {
+        useStrictHideStyleTag?: boolean;
+        rules: Array<{
```
laghee, line 6: Since we're using this same `rules` structure twice, this might be a good place to reduce repetition and refactor to make a generalized `ElementHidingRule` (or whatever), similar to how we have [DescriptorAPIChange](https://github.com/duckduckgo/privacy-configuration/blob/8448502c97a3a2ef63d35932a7f11980a632a8c2/schema/features/api-manipulation.ts#L7) that's then used in [FullAPIManipulationOptions](https://github.com/duckduckgo/privacy-configuration/blob/8448502c97a3a2ef63d35932a7f11980a632a8c2/schema/features/api-manipulation.ts#L16).

schema/features/element-hiding.ts
icodebyamanda: @laghee I discussed with Dave and both are actually needed. Change was submitted

schema/features/element-hiding.ts
jonathanKingston: I'd fast follow this. But agreed. (especially as the next step is to import these types back into the code, so naming thinking about how they'll be used is important... you may want to tackle these together as a next step).

schema/features/element-hiding.ts
jonathanKingston: What does removing this  `[key: string]: any;` do?
Do tests fail?

Again I would fast follow with trying to remove this, but essentially it's currently an open contract for more keys.

jonathanKingston (approved):
I'd merge this and follow up with some of @laghee's suggestions.

Tests pass, so worst case scenario we're:
- Too restrictive on permitted keys, that's fine as it's unlikely to end in incident, someone adding an unsupported rule can change the type.
- In the case of `[key: string]: any;` we're allowing any key extensions, that risks doing accidental things in future. I think we should follow up with locking that down.

schema/features/element-hiding.ts
dharb: @icodebyamanda sorry I meant we needed both hideTimeouts and unhideTimeouts, didn’t realize this was in reference to this [key:string]. I would just do as Jonathan says and try removing it and seeing what fails, then add whatever keys are missing.

schema/features/element-hiding.ts
icodebyamanda: I'll follow up on this one @laghee

PR #3967, opened 2025-10-22
Add vpnMenuItem sub-feature on iOS

rachelmcr:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/task/1211694500482776

## Description

Adds the vpnMenuItem feature on iOS targeting subscription-eligible users in the US and UK.

### Feature change process:

---

> [!NOTE]
> Adds `privacyPro.features.vpnMenuItem` (disabled) targeting `GB` and `US` in `overrides/ios-override.json`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 835218b4415a236a47c140494a54b7437c9f7aa4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/overrides/ios-override.json
+++ b/overrides/ios-override.json
@@ -1417,12 +1417,10 @@
                     "state": "disabled",
                     "targets": [
                         {
-                            "localeCountry": "gb",
-                            "isPrivacyProEligible": true
+                            "localeCountry": "GB"
                         },
                         {
-                            "localeCountry": "us",
-                            "isPrivacyProEligible": true
+                            "localeCountry": "US"
                         }
                     ]
                 }


aataraxiaa (approved):
LGTM @rachelmcr. Unsure if this Cursor comment is accurate but worth checking - https://github.com/duckduckgo/privacy-configuration/pull/3967#discussion_r2451411041.

rachelmcr:
Ah, hm. I copied the config from Android thinking it would work the same way on iOS, but on closer inspection:

* We do an exact string match for the `localeCountry` in the app, so those should be capitalized.
* It looks like on Apple we don’t support the `isPrivacyProEligible` target. We define `PrivacyConfigurationData.PrivacyFeature.Feature.Target` and only support `localeCountry` and `localeLanguage`. I’ll move that check into the feature code rather than relying on the config.

rachelmcr:
@aataraxiaa Since I made significant changes, could you take another look? I moved the subscription eligibility check into the client as part of https://github.com/duckduckgo/apple-browsers/pull/2263, so the config now just checks the country.

PR #4224, opened 2025-12-09
Add js user agent client hints brand header-matching feature

laghee:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/45878998844068/task/1212017506660685?focus=true

Not sure if we want to flip to `enabled` immediately or do that separately? Wdyt, @jonathanKingston ? (looks like we got the native bump to C-S-S 12.2.0 into preview v.0.141.5 🎉 )

### Feature change process:

---

> [!NOTE]
> Introduces the `uaChBrands` feature (schema, base config, build integration) and enables it on Windows with a conditional `additionalCheck`; also adds `preview` to condition blocks.
> 
> - **Features**:
>   - **`uaChBrands`**:
>     - Adds schema (`schema/features/ua-ch-brands.ts`) and base config (`features/ua-ch-brands.json`).
>     - Wires into config types (`schema/config.ts`).
>     - Propagates temp unprotected exceptions in build (`index.js`).
>     - Windows override: enabled with `additionalCheck` and conditional patch when `internal` and `preview` (`overrides/windows-override.json`).
> - **Schema/Core**:
>   - Adds `preview` condition support to `ConditionBlock` (`schema/feature.ts`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3aa170bdbec4fbc6434dfff4d8dcfc7c43c20cce. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/schema/feature.ts
+++ b/schema/feature.ts
@@ -80,7 +80,7 @@ type CSSInjectFeatureSettingsPatches = {
         patchSettings: Operation<JSONValidValueType>[];
     }[];
     conditionalChanges?: {
-        condition: ConditionBlock | Array<ConditionBlock>;
+        condition: ConditionBlock | ConditionBlock[];
         patchSettings: Operation<JSONValidValueType>[];
     }[];
 };


overrides/windows-override.json
```diff
@@ -2545,6 +2545,9 @@
         "clientBrandHint": {
             "state": "enabled"
         },
+        "uaChBrands": {
+            "state": "enabled"
```
jonathanKingston: We could use something like this if you wanted to limit damage:
```
                "additionalCheck": "disabled",
                "conditionalChanges": [
                    {
                        "condition": [
                            {
                                "internal": true
                            },
                            {
                                "preview": true
                            },
                        ],
                        "patchSettings": [
                            {
                                "op": "replace",
                                "path": "/additionalCheck",
                                "value": "enabled"
                            }
                        ]
                    },
```

overrides/windows-override.json
```diff
@@ -2545,6 +2545,38 @@
         "clientBrandHint": {
             "state": "enabled"
         },
+        "uaChBrands": {
+            "state": "enabled",
+            "settings": {
+                "additionalCheck": "disabled",
+                "conditionalChanges": [
+                    {
+                        "condition": {
```
jonathanKingston: Did the array format not work here? (I am 99% sure it should)

jonathanKingston: Yeah should work: https://github.com/duckduckgo/privacy-configuration/blob/f0004f26a0e1c1b476c927a0a0f6092e42372b85/schema/feature.ts#L82 lowers the copy and paste but also happy without making that change.

laghee: Aha, I was worried it wouldn't work and I saw you had added the `pageContext` one, so I erred on the side of caution. Can def merge them.

overrides/windows-override.json
```diff
@@ -2545,6 +2545,26 @@
...
         },
+        "uaChBrands": {
+            "state": "enabled",
+            "settings": {
+                "additionalCheck": "disabled",
+                "conditionalChanges": [
+                    {
+                        "condition": [
+                            { "internal": true },
+                            { "preview": true }],
```
jonathanKingston, line 2556: I love that prettier is the most pedantic tool alive, but just lets folk do this. (I don't care really but just 😆 ).

laghee: Damn, I didn’t see this. Don’t we have a rule for this actually? I thought @kzar added one, but maybe it was allowlist specific 😢

PR #5332, opened 2026-06-11
Harden MSP stale-exemption workflow against misleading PRs

jonathanKingston:
## Summary

The `msp-filter-stale.yml` workflow can open PRs whose diff does not match their title, and it repeatedly tries to remove a permanent test exception. This PR fixes both.

### 1. Misleading auto-generated PRs (e.g. #5325)

#5325 was titled "Remove stale exemptions..." but contained only an unrelated `package-lock.json` metadata change, versus the well-formed #3964. Two root causes:

- **Silent failures slipped past the guard.** `scripts/filter-malicious.js` catches its own errors (`main().catch(console.error)`) and still exits `0`, so a crash produces empty stdout. The old guard only bailed on the literal string `No updates`, so empty output proceeded to PR creation with an empty `PR_ID` (hence the `remove-stale-exemptions-` branch with no ID suffix on #5325). Now the job fails loudly when no `PR ID:` line is emitted, and PR creation is gated on a step output.
- **`add-paths: ./` swept in unrelated churn.** Combined with `npm i` / `npm run build` / `npm run lint-fix` mutating the working tree, any side effect (e.g. the `package-lock.json` `from` rewrite) got committed under the "Remove stale exemptions" title. `add-paths` is now scoped to the config files this workflow actually edits.

Also added `set -euo pipefail` for stricter shell error handling.

### 2. Retain the `broken.third-party.site` test exception (requested in #3964)

`broken.third-party.site` is a synthetic test page used to verify that exceptions work correctly. It is never present in the malicious dataset, so the staleness check flagged it for removal on every run (#3964). `filter-malicious.js` now treats it as a permanent exception that is always retained.

## Notes

- No active security vulnerability existed in either prior PR's diff; the workflow change is a provenance/integrity and robustness hardening so auto-generated PRs only ever contain the intended MSP config changes.
- Scoped override files (`ios`, `macos`, `windows`, `android`) plus `features/malicious-site-protection.json` match the `platforms`/`defaultConfig` written by `filter-malicious.js` (extension is intentionally not processed).

## Test plan

- [ ] No stale exceptions → job succeeds, no PR.
- [ ] Simulated script error / empty output → job fails, no PR opened.
- [ ] Real stale exceptions → PR contains only the MSP config files, with a populated `PR ID` and branch suffix.
- [ ] `broken.third-party.site` is never proposed for removal.

### DIFF
--- a/.github/workflows/msp-filter-stale.yml
+++ b/.github/workflows/msp-filter-stale.yml
@@ -14,35 +14,56 @@ jobs:
                   repository: duckduckgo/privacy-configuration
                   path: privacy-configuration/
             - name: Execute Update Script
+              id: prep
               run: |
+                  set -euo pipefail
                   cd ./privacy-configuration/
-                  npm i 
+                  npm i
                   npm run build
                   DATE="$(date -u +'%Y-%m-%d')"
-                  echo "DATE=$DATE" >> $GITHUB_ENV
+                  echo "DATE=$DATE" >> "$GITHUB_ENV"
                   TEMPLATE="$(node ./scripts/filter-malicious.js)"
-                  # If TEMPLATE contains 'No updates' then exit with success
-                  if [[ $TEMPLATE == *"No updates"* ]]; then
+                  # No stale exceptions to remove: nothing to do, skip PR creation.
+                  if [[ "$TEMPLATE" == *"No updates"* ]]; then
                     echo "No updates found. Exiting with success."
+                    echo "should_create=false" >> "$GITHUB_OUTPUT"
                     exit 0
-                  fi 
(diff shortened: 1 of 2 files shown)


daxtheduck:
### [JSON approval analysis](https://github.com/duckduckgo/privacy-configuration/blob/main/.github/workflows/auto-respond-pr.yml#L5-L12):
*Time updated:* Thu, 11 Jun 2026 01:16:32 GMT
<details open>
<summary>latest</summary>

</details>

## 🎯 OVERALL APPROVAL STATUS
**✅ AUTO-APPROVED**

